PR #13, opened 2020-09-03
Updates to page names and button text

jesseyuen:
We now refer to 'Records' as 'Applications' in url path and adding a 'record' becomes 'trainee' in public facing UI like buttons.


jesseyuen:
Please excuse the spelling mistake in the branch name! 🙈

PR #22, opened 2020-09-08
Start page added

jesseyuen:


app/views/start-page.html
```diff
@@ -1,31 +1,66 @@
...
+    <div class="govuk-width-container">
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+
+          <h1 class="start-page-banner__text start-page-banner__heading govuk-heading-xl">Register trainee teachers</h1>
+          <p class="start-page-banner__text start-page-banner__description govuk-body govuk-!-font-size-27">A service for collecting and maintaining trainee teacher registration data.</p>
+          <div class="start-page-banner__actions">
+            {{ govukButton({
+              text: "Get started",
+              href: "/records.html",
```
edwardhorsford:
```suggestion
              href: "/records",
```

We don't need to provide an extension here. I've now removed similar from other buttons.

app/views/start-page.html
```diff
@@ -1,31 +1,66 @@
...
+    <ul class="govuk-list govuk-list--bullet">
+      <li>upload Assessment Only (AO) candidate data</li>
+      <li>get a Teacher Reference Number (TRN) for AO candidates</li>
+      <li>recommend candidates for Qualified Teacher Status (QTS)</li>
+      <li>check and update your AO candidate datas</li>
+    </ul>
+
+    <p class="govuk-body">If you are a Higher Education Institute (HEI), you must submit your AO data using the Register trainee teacher service. You do not submit it via the Higher Education Statistics Agency (HESA).</p>
+    
+    <p class="govuk-body">You can only register trainee teachers for the current academic year. If you need to register a trainee teacher within a different academic year contact <a href="mailto:itt.datamanagement@education.gov.uk">itt.datamanagement@education.gov.uk</a>.</p>
```
edwardhorsford: should this have `govuk-link` on the mailto?

edwardhorsford (changes requested):

app/views/start-page.html
jesseyuen: Good shout

app/views/start-page.html
jesseyuen: There seems to be generic styles to allow for links with no classes but I suspect it's more robust wiht the class included.

edwardhorsford (approved):

PR #45, opened 2020-09-16
Guidance update

jesseyuen:
For this round of research Greg is keen to include the AO guidance and data requirements at the start of the flow. This PR to update the start page reflects such desires.

I don't think the table in `/data-requirements` would scale to 12 routes but I do think there is perhaps something in a comparison table for looking at what is common and what is unique. 🤷‍♂️


app/views/guidance.html
```diff
@@ -9,35 +19,17 @@
...
         <h2 class="govuk-heading-m">Assessment Only route</h2>
         <p class="govuk-body">AO is a route to Qualified Teacher Status (QTS) in England for:</p>
         <ul class="govuk-list govuk-list--bullet">
           <li>experienced teachers with a degree</li>
           <li>people with teaching qualifications from other countries</li>
         </ul>
         <p class="govuk-body">It allows teachers to demonstrate that they already meet all the QTS standards without the need for further training.</p>
-        <p class="govuk-body">Find out more about <a href="https://www.gov.uk/government/publications/the-assessment-only-route-to-qts#:~:text=The%20assessment%20only%20route%20to%20qualified%20teacher%20status%20(%20QTS%20)%20allows,an%20accredited%20and%20approved%20provider"> AO route policy</a>.</p>
+        <p class="govuk-body">Find out more about the <a href="https://www.gov.uk/government/publications/the-assessment-only-route-to-qts" class="govuk-link"> AO route policy</a>.</p>
```
edwardhorsford:
```suggestion
        <p class="govuk-body">Find out more about the <a href="https://www.gov.uk/government/publications/the-assessment-only-route-to-qts" class="govuk-link">AO route policy</a>.</p>
```

app/views/data-requirements.html
```diff
@@ -0,0 +1,169 @@
...
+            "
+          },
+          {
+            html: ""
+          }
+        ],
+        [
+          {
+            html: "
+            <h2 class='govuk-heading-s'>Diversity information</h2>
```
edwardhorsford: I think this will need to be more nuanced - as this information *isn't* required.

edwardhorsford (approved):

app/views/data-requirements.html
jesseyuen: That's a good shout. Will loop back with Greg.

jesseyuen: **Diversity information (optional)** might be enough in this instance.

PR #78, opened 2020-10-16
Add conditionals for ethnicity and disabilities

edwardhorsford:
The two diversity sections should both support users being able to answer 'other' and provide a freetext response. I didn't have time to build this originally, so adding now. These mostly follow the patterns from Apply but with some text tweaks to refer to `their` rather than `your`.

## Ethnic background

Optionally collecting another ethnic background:
Screenshot 2020-10-16 at 11 44 35

Optionally displaying the ethnic background:
Screenshot 2020-10-16 at 11 44 42

## Disabilities

Optionally collecting a different disability:
Screenshot 2020-10-16 at 11 46 16

Optionally displaying the disability:
Screenshot 2020-10-16 at 11 47 08

### DIFF
--- a/app/views/_includes/forms/ethnic-background.html
+++ b/app/views/_includes/forms/ethnic-background.html
@@ -1,50 +1,50 @@
 
 {% set ethnicGroup = record.diversity.ethnicGroup %}
 
-{# {% set asianOtherHtml %}
+{% set asianOtherHtml %}
   {{ govukInput({
     classes: "govuk-!-width-two-thirds",
     label: {
-      text: "Your Asian background (optional)"
+      text: "Describe their Asian background (optional)"
     }
-  } | decorateApplicationAttributes(["equality-monitoring", "ethnic-background-other"])) }}
+  } | decorateAttributes(record, "record.diversity.ethnicBackgroundOther"))}}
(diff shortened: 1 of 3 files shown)


app/views/_includes/forms/ethnic-background.html
```diff
@@ -1,50 +1,50 @@
...
-  } | decorateApplicationAttributes(["equality-monitoring", "ethnic-background-other"])) }}
+  } | decorateAttributes(record, "record.diversity.ethnicBackgroundOther"))}}
 {% endset -%}
 
 {% set mixedOtherHtml %}
   {{ govukInput({
     classes: "govuk-!-width-two-thirds",
     label: {
-      text: "Your Mixed background (optional)"
+      text: "Describe their Mixed background (optional)"
```
jesseyuen, line 26: Should 'Mixed' be capitalised in this instance?

edwardhorsford: It's following the [design system's example casing](https://design-system.service.gov.uk/patterns/ethnic-group/#mixed-or-multiple-ethnic-groups) - which I think follows ONS' example.

jesseyuen (approved):

PR #98, opened 2020-10-30
Add support for sorting records

edwardhorsford:
Adds support for sorting records.

As a byproduct I've also moved the search input to the filter bar - which matches Manage.

Screenshot 2020-11-03 at 11 24 01


app/assets/sass/components/_all-records-search.scss
```diff
@@ -17,4 +17,91 @@
...
+
+.gem-c-search__submit {
+  border: 0;
+  cursor: pointer;
+  border-radius: 0;
+  position: relative;
+  padding: 0;
+  width: $input-size;
+  height: $input-size;
+  background-image: url("/public/images/search-button.png");
```
jesseyuen, line 37: I feel like an inline <svg> icon with the appropriate A11Y support (`aria-hidden="true" focusable="false"`) would be a better option than a png background image? Fine for the prototype though.

app/assets/sass/components/_all-records-search.scss
```diff
@@ -17,4 +17,91 @@
...
+  background-repeat: no-repeat;
+  background-position: 2px 50%;
+  text-indent: -5000px;
```
jesseyuen, line 40: Wrapping the button text in a `<span class="govuk-visually-hidden">Search</span>` would remove the need for the extreme text-indent value.

jesseyuen (approved):

app/assets/sass/components/_all-records-search.scss
edwardhorsford: This stuff was copied from the govuk publishing app. I've now commented this bit out though.

app/assets/sass/components/_all-records-search.scss
edwardhorsford: This stuff was copied from the govuk publishing app. I've now commented this bit out though.

PR #99, opened 2020-10-30
lots of content changes and a few formatting changes

EmmaFrith:
I wasn't able to mirror the review page with the task list

### DIFF
--- a/app/views/home.html
+++ b/app/views/home.html
@@ -2,13 +2,18 @@
 
 {% block content %}
 
-    <h1 class="govuk-heading-l">{{serviceName}}</h1>
-      <p class="govuk-body govuk-!-margin-bottom-6">Register trainees with the Department for Education and record the outcome of their training.</p>
-    {{ govukButton({
-      text: "Add a trainee",
-      href: "./new-record/new",
-      isStartButton: true
-    }) }}
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds-from-desktop">
+        <h1 class="govuk-heading-l">{{serviceName}}</h1>
+        <p class="govuk-body-l">Register trainees with the Department for Education and record the outcome of their training.</p>
+        {{ govukButton({
+          text: "Add a trainee",
+          href: "./new-record/new",
+          isStartButton: true
+        }) }}
+      </div>
+    </div>
+
 
     <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
 
--- a/app/views/home.html
+++ b/app/views/home.html
@@ -3,7 +3,7 @@
 {% block content %}
 
     <h1 class="govuk-heading-l">{{serviceName}}</h1>
-    <p class="govuk-body">Collect and manage trainee teacher registration data.</p>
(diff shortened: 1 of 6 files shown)


app/views/home.html
```diff
@@ -3,7 +3,7 @@
 {% block content %}
 
     <h1 class="govuk-heading-l">{{serviceName}}</h1>
-    <p class="govuk-body">Collect and manage trainee teacher registration data.</p>
+      <p class="govuk-body govuk-!-margin-bottom-6">Register trainees with the Department for Education and record the outcome of their training.</p>
```
edwardhorsford: I don't think this should be indented.

How about using `govuk-body-l` for this? would make it a bit bigger and maybe negate the need for the extra margin-bottom.

app/views/new-record/submitted.html
```diff
@@ -28,25 +28,17 @@
...
-  <p class="govuk-body">You can recommend the trainee for QTS once their TRN is received and they’ve met the requirements.</p>
-{% endif %}
-
-<p class="govuk-body">Would you like to <a href="/new-record/new" class="govuk-link">add another trainee</a>?</p>
-
-{# {{ govukButton({
-  text: "View record",
-  href: "/record/" + data.recordId
-}) }}
+<p class="govuk-body govuk-!-margin-bottom-7">Keep the trainee record up to date and record the outcome of their training, even if you do not recommend them for a teacher status.</p>
```
edwardhorsford: Unsure about `a teacher status`.

app/views/new-record/overview.html
```diff
@@ -74,11 +74,20 @@ <h2 class="govuk-heading-m">Trainee details</h2>
...
+  } if record | requiresSection("diversity"),
+  {
+    text: "Degree",
+    href: "degree/confirm" if (record.degree.items | length) else "degree/add",
+    id: "qualification-degree",
+    tag: {
+      classes: record.degree | getStatusText | getStatusClass,
+      text: record.degree | getStatusText
+    }
+  } if record | requiresSection("degree")
```
edwardhorsford: What's your thinking for moving this?

app/views/new-record/submitted.html
EmmaFrith: I wasn't sure how to make this conditional on whether its EYTS or QTS - hence the vagueness! can you help me do that?

app/views/new-record/overview.html
EmmaFrith: I was anticipating the direction of the service on the basis of UR. 

1. providers told us they often enter trainees which are all on the same route + course - meaning that only thing that's different each time is the candidates' unique information, like their degree info, diversity info and address etc. 

if we eventually make it possible to apply a route and programme across multiple candidates, it makes sense to split this page up into 'about the training' (consistent across multiple trainees) and 'about the candidate' (unique information each time). 

2. we got some feedback that providers want to skip quickly through the sections, and wouldn't spend much time on this task list page

essentially, splitting it up into too many sections which don't correspond to the way providers think/work doesn't feel necessary

app/views/home.html
EmmaFrith: I'm not quite sure what I did here - help pls!

edwardhorsford (approved):
Looks great! 🎉

PR #105, opened 2020-11-05
Update degree page

edwardhorsford:
Moves collecting the degree type from the initial page to the subsequent page.

This is easier for us to develop, but also might help with an issue where users attempt to include a subject with the degree type (BSC Biology, etc).

Also took a bit of time to add some helpers for setting the widths of inputs and autocompletes (copied from my [last project here](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/app/assets/stylesheets/helpers/field-widths.scss)). In a later PR I'll go back and adjust other inputs to match.

Screenshot 2020-11-05 at 17 37 26
Screenshot 2020-11-05 at 17 37 34

### DIFF
--- a/app/assets/sass/application.scss
+++ b/app/assets/sass/application.scss
@@ -42,9 +42,11 @@ $govuk-assets-path: '/govuk/assets/';
 @import "overrides/moj-pagination";
 @import "overrides/tabs";
 @import "overrides/dash-list";
+@import "overrides/summary-card";
 
 // Misc
 @import "helpers";
+@import "helpers/field-widths";
 
 // Add extra styles here, or re-organise the Sass files in whichever way makes most sense to you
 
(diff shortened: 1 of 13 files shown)


EmmaFrith:
perhaps going off piste from this PR but:

after adding an undergrad degree, if you select 'add another degree', the next page heading is still 'Add undergraduate degree' - can we change the content for subsequent degrees to 'Add a degree' as we don't know whether they'll be adding a postgrad or an undergrad?

@edwardhorsford

edwardhorsford:
> perhaps going off piste from this PR but:
> 
> after adding an undergrad degree, if you select 'add another degree', the next page heading is still 'Add undergraduate degree' - can we change the content for subsequent degrees to 'Add a degree' as we don't know whether they'll be adding a postgrad or an undergrad?
> 
> @edwardhorsford

Good idea @EmmaFrith - now included in this PR.

jesseyuen (approved):

PR #107, opened 2020-11-06
A few small content changes 

EmmaFrith:
a few small content changes 

- to home page and linked sections to make page headings consistent with link content 
- to data requirements page to make language more human and reduce content length 

@edwardhorsford I couldn't figure out how to add text into a list on the data requirements page, so have formatted the text as a list item for now

### DIFF
--- a/app/views/data-requirements.html
+++ b/app/views/data-requirements.html
@@ -2,7 +2,7 @@
 {% extends "_templates/_page.html" %}
 {% set backLink = '/home' %}
 
-{% set pageHeading = 'Data requirements' %}
+{% set pageHeading = 'Check what data you need' %}
 
 {% block content %}
 
@@ -15,7 +15,7 @@
       <li><a class="govuk-link" href="#section-all-routes">Data required for all routes</a></li>
       <li><a class="govuk-link" href="#section-assessment-only">Specific data for Assessment Only</a></li>
       <li><a class="govuk-link" href="#section-provider-led">Specific data for Provider-led</a></li>
-    </ul> 
+    </ul>
   </div>
 </div>
 
@@ -190,7 +190,7 @@
             {
               html: "
               <h2 class='govuk-heading-s'>This is a placeholder for Provider-led route data</h2>
-              
+
               "
             }
           ]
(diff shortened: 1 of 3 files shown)


app/views/data-requirements.html
```diff
@@ -146,7 +148,7 @@ <h2 class='govuk-heading-s'>Non-UK degrees</h2>
         captionClasses: "govuk-visually-hidden",
         head: [
           {
-            text: "Specific data for Assessment Only",
+            text: "Specific data for assessment-only trainees",
```
edwardhorsford:
```suggestion
            text: "Specific data for assessment-only candidates",
```

I wonder if in this case we should use the correct term

app/views/data-requirements.html
```diff
@@ -2,20 +2,18 @@
...
     <h1 class="govuk-heading-l">{{pageHeading}}</h1>
-    <p class="govuk-body">We only ask for the data we require to register trainee teachers. We have outlined what you need to collect for each route.</p>
-    <h2 class="govuk-heading-m">Contents</h2>
     <ul class="govuk-list govuk-list--dash govuk-!-margin-bottom-8">
-      <li><a class="govuk-link" href="#section-all-routes">Data required for all routes</a></li>
-      <li><a class="govuk-link" href="#section-assessment-only">Specific data for Assessment Only</a></li>
-      <li><a class="govuk-link" href="#section-provider-led">Specific data for Provider-led</a></li>
-    </ul> 
+      <li><a class="govuk-link" href="#section-all-routes">Standard data for all trainees</a></li>
+      <li><a class="govuk-link" href="#section-assessment-only">Specific data for assessment-only trainees</a></li>
```
edwardhorsford: Should we call them candidates here?

app/views/data-requirements.html
```diff
@@ -81,7 +79,11 @@ <h2 class='govuk-heading-s'>Contact details</h2>
             {
               html: "
               <h2 class='govuk-heading-s'>Diversity information (optional)</h2>
-              <ul class='govuk-list'>
+              <ul class='govuk-list govuk-list--spaced'>
+                <li>
+                  If the trainee chose not to share diversity information during the application process, you should not ask them for it for the prupose of registering them.
+                </li>
+                <ul class='govuk-list'>
```
edwardhorsford: I don't think this ul should be here

edwardhorsford (changes requested):

edwardhorsford (approved):

app/views/data-requirements.html
EmmaFrith: yes!

edwardhorsford (approved):

PR #111, opened 2020-11-11
New header

edwardhorsford:
Updates our header to be more consistent with manage and support adding new sections

Screenshot 2020-11-11 at 17 03 17

As an easter egg there's also an alternate design using a dark header. Talking with other teams in DFE there wasn't much appetite to change to it, so it's disabled for now.


app/views/layout.html
```diff
@@ -56,38 +59,77 @@
...
+  {% set navItems = [{
+      text: 'Home',
+      href: '/home',
+      active: true if navActive == 'home'
+    }, {
+      text: 'Trainee records',
+      href: '/records',
+      active: true if navActive == 'records'
+    }, {
+      text: 'Courses',
```
jesseyuen, line 92: Do we need Courses?

edwardhorsford: It's disabled for now.

jesseyuen (approved):

fofr:
Why is the phase bar below the nav, rather than matching Manage applications and the Design system by putting it above the nav?

edwardhorsford:
> Why is the phase bar below the nav, rather than matching Manage applications and the Design system by putting it above the nav?

@fofr see conversation in #twd_bat_design - we're trying to avoid the odd sandwich effect of the phase banner placed between nav and black bar.

PR #115, opened 2020-11-13
Adding the declaration page (WIP)

jesseyuen:
A placeholder for the Declaration page. It sounds like there is some question about it's necessity. Parking the page for now behind a feature flag.

This PR includes the `hidePrimaryNav` flag which is used in [#116](https://github.com/DFE-Digital/register-trainee-teachers-prototype/pull/116)
 
Screenshot 2020-11-17 at 10 07 58

PR #116, opened 2020-11-13
Move phase banner within blue bar, and make nav conditional

edwardhorsford:
Moves the phase banner within the blue bar.

The page also sets a variable to turn off the primary nav - but the condition to actually remove it is part of #115 

Screenshot 2020-11-13 at 11 33 47

### DIFF
--- a/app/views/start-page.html
+++ b/app/views/start-page.html
@@ -1,6 +1,7 @@
-{% extends "layout.html" %}
+ {% extends "layout.html" %}
 {% set bodyClasses = "start-page" %}
 {% set hideNav = true %}
+{% set hidePrimaryNav = true %}
 
 {% block pageTitle %}
   {{ pageHeading }} - GOV.UK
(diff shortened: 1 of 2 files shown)


app/views/start-page.html
```diff
@@ -1,28 +1,33 @@
...
-    {% endblock %}
-  </div>
   <div class="start-page-banner">
 
     <div class="govuk-width-container">
+      <div class="govuk-phase-banner">
+        <p class="govuk-phase-banner__content">
+          {{ govukTag({
+            text: "prototype",
+            classes: "govuk-phase-banner__content__tag start-page-banner__tag"
```
jesseyuen, line 24: `indent(4) | trim` - what do these do @edwardhorsford ?

edwardhorsford: It's stuff copied from the deisgn system. When you call the phase banner component, inside it's just calling the `govukTag` component with these options. From the look of it, I'd assume it indents some content and then trims whitespace from it.

jesseyuen (approved):

PR #118, opened 2020-11-13
Add initial support for hints inside autocompletes

edwardhorsford:
Adds some hacky support for a hint inside the results of the autocomplete.

To use, pass a name in the form `Text | hint` - the hint will be split off and shown in grey.

Screenshot 2020-11-13 at 13 46 51


jesseyuen:
In Find/Pubish the pattern for writing course name and code is `Art and design (2P35)`. Wondering if we should be consistent?

jesseyuen (approved):

PR #119, opened 2020-11-17
Initial support for provider courses

edwardhorsford:
Adding backend data support for publish data.

Key changes:
* New programme details generator to spit out publish or non publish data
* Generate fake courses for each provider
* Seed records use publish courses 50% of the time - though as publish courses use a different route, these don't appear unless you enable other routes.


app/data/training-route-data.js
```diff
@@ -32,7 +32,7 @@ let extraRoutes = [
   "Early years - grad emp",
   "Early years - grad entry",
   "Early years - assessment only",
-  "Early years undergraduate",
+  "Early years - undergraduate",
   "School Direct salaried",
   "School direct tuition fee",
```
jesseyuen, line 37: Discrepancy in capitalisation on these two lines. Is it Direct or direct?

jesseyuen (approved):

app/data/training-route-data.js
edwardhorsford: Went with lowercase for now. Good catch.

PR #124, opened 2020-11-18
Add stub sign-in page

edwardhorsford:
Adds a sign-in page which would then link to DfE sign-in

### DIFF
--- a/app/views/sign-in.html
+++ b/app/views/sign-in.html
@@ -20,7 +20,7 @@
       href: 'declaration' if data.settings.includeDeclaration else 'home'
     }) }}
 
-    <p class="govuk-body-m">If you do not have a DfE Sign-in account but need access to Register trainee teachers, email registerateacher@digital.education.gov.uk.</p>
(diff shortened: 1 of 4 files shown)


app/views/sign-in.html
```diff
@@ -0,0 +1,25 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+
+    <h1 class="govuk-heading-l">{{pageHeading}}</h1>
+
+    <p class="govuk-body-l">You must sign in to your account to register trainee teachers.</p>
+
+    {{ govukButton({
+      "text": "Sign in using DfE Sign-in",
+      href: 'declaration' if data.settings.includeDeclaration else 'home'
```
jesseyuen, line 20: I woud assume you'd lead with the `if` like this:

`if data.settings.includeDeclaration 'declaration' else 'home'`

Interesting that it works in this order!

jesseyuen: Actually, my way breaks it. Nice.

jesseyuen (approved):

EmmaFrith:
have added a couple of commits @edwardhorsford 

Screenshot 2020-11-19 at 11 53 47

PR #143, opened 2020-11-26
Let trainee start date be edited on records

edwardhorsford:
#117 didn't let users edit the trainee start date on existing records. This adds it in.

Screenshot 2020-11-26 at 12 41 53

Also updates a bunch of form actions to more consistently use referrers - which we'll defer to when deciding where to return a user to after going down a flow.


app/views/_includes/forms/trainee-start-date.html
```diff
@@ -0,0 +1,42 @@
...
+      },
+      {
+        name: "year",
+        classes: "govuk-input--width-4",
+        value: commencementStartDateArray["2"]
+      }
+    ]
+}) }}
+
+<input type="hidden" name="record[trainingDetails][traineeStarted]" value="true">
```
jesseyuen, line 36: `<input type="hidden" name="record[trainingDetails][traineeStarted]" value="true">`

What's this doing? ...just curious!

edwardhorsford: This manually sets some data in the auto-data store.

On the 'add' journey, we first have a question 'have they started yet?' and then only show the date if the answer to that is yes.

The question didn't seem relevant once you're editing an existing record - so at the point you set a date, I'm manually setting the answer to that question.

jesseyuen (approved):

PR #157, opened 2020-12-04
Update publish pages

edwardhorsford:
This PR splits out asking about publish courses from asking about route. It prioritises publish courses as using them means we get more accurate data and save users filling in information we already have. The other approach was ok whilst we support a low number of routes, but got really busy when we start to support all 12 routes.

If a provider doesn't have any publish courses, we can skip the `pick-course` page and go straight to the `pick-route` page - which is similar to before.

The `pick-course` page is also updated to deal with _lots_ of courses. When the number of courses is over `x` we can show an autocomplete instead.

## Pick course with low number of courses
Screenshot 2020-12-04 at 17 38 46

## Pick course with high number of courses
Screenshot 2020-12-04 at 17 39 17

## Pick route
Screenshot 2020-12-04 at 17 38 13

This should hopefully be fully working. You can go in and edit existing details and change course.


app/data/session-data-defaults.js
```diff
@@ -29,6 +29,7 @@ let settings = {}
 // Simplify structure so it can be worked with from admin page
 settings.enabledTrainingRoutes = Object.values(trainingRoutes).filter(route => route.defaultEnabled == true).map(route => route.name).sort()
 settings.includeTimeline = 'true'
+settings.courseLimit = 50
```
jesseyuen: What does this do?

app/lib/record.js
```diff
@@ -37,3 +37,10 @@ exports.requiresSection = (record, sectionName) => {
   return requiredSections.includes(sectionName)
 }
 
+exports.sortAlphabetical = (x, y) => {
```
jesseyuen, line 41: Could do with a comment explainign what the function does.

app/routes/shared-edit-routes.js
```diff
@@ -30,40 +30,62 @@ module.exports = router => {
...
 
-  // Picking a course or a route
-  router.post(['/:recordtype/:uuid/programme-details','/:recordtype/programme-details'], function (req, res) {
+  // Show pick-course or pick-route depending on if current provider has courses
+  // on Publish
+  router.get(['/:recordtype/:uuid/programme-details','/:recordtype/programme-details'], function (req, res) {
+    const data = req.session.data
+    let recordPath = utils.getRecordPath(req)
+    let referrer = utils.getReferrer(req.query.referrer)
+    let providerCourses = data.courses["University of Southampton"].courses
```
jesseyuen, line 39: Why is "University of Southampton" is specified here Ed?

app/views/_includes/forms/programme-details/pick-course.html
```diff
@@ -1,65 +1,194 @@
...
+
+  {% set courseItems = [] %}
+
+  {% for course in providerCourses %}
+    {% set courseItems = courseItems | push({
+      text: course.subject + " (" + course.code + ")",
+      value: course.id,
+      checked: checked(selectedCourse, course.id),
+      hint: {
+        _text: course.route + " — " + course.summary,
```
jesseyuen: Consider deleting `_text: course.route + " — " + course.summary,` if you're not using it?

app/views/admin.html
```diff
@@ -48,6 +48,17 @@ <h1 class="govuk-heading-l">{{pageHeading}}</h1>
         ]
       } | decorateAttributes(data, "data.settings.includeDeclaration")) }}
 
+      {{ govukInput({
+        label: {
+          text: "Number of publish courses provider offers",
```
jesseyuen:
```suggestion
          text: "Number of publish courses the provider offers",
```

app/data/session-data-defaults.js
edwardhorsford: Have added a comment in the code to make it clearer. Is a hacky way of letting us set the number of Publish courses that a provider offers.

app/lib/record.js
edwardhorsford: Added a comment to it.

jesseyuen (approved):

PR #160, opened 2020-12-07
Delete draft confirmation added 

jesseyuen:
Screenshot_2020-12-07 Delete this draft record - Register trainee teachers - GOV UK(1)

### DIFF
--- a/app/views/_templates/_new-record.html
+++ b/app/views/_templates/_new-record.html
@@ -67,8 +67,8 @@
         {% if returnLink %}
           <p class="govuk-body"><a href="{{returnLink.href}}" class="govuk-link">{{returnLink.text}}</a></p>
         {% endif %}
-        {% if cancelLink %}
-          <p class="govuk-body"><a href="{{recordPath}}" class="govuk-link">Cancel and return to record</a></p>
+        {% if returnLink %}
+          <p class="govuk-body"><a href="{{returnLink.href}}" class="govuk-link">{{returnLink.text}}</a></p>
         {% endif %}
 
         {# {% if not hideReturnLink %}
--- a/app/views/_templates/_new-record.html
+++ b/app/views/_templates/_new-record.html
@@ -67,6 +67,9 @@
         {% if returnLink %}
           <p class="govuk-body"><a href="{{returnLink.href}}" class="govuk-link">{{returnLink.text}}</a></p>
         {% endif %}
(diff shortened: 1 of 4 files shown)


app/views/new-record/delete-draft/confirm.html
```diff
@@ -0,0 +1,27 @@
+{% extends "_templates/_record-form.html" %}
```
edwardhorsford: This is the template for existing records, not draft ones. When swapping over, check there's nothing else we're missing.

app/views/new-record/delete-draft/confirm.html
```diff
@@ -0,0 +1,27 @@
...
+
+{% set pageHeading = "Delete this draft record" %}
+{% set backLink = './../overview' %}
+{% set backText = "Back to draft record" %}
+{% set gridColumn = 'govuk-grid-column-full' %}
```
edwardhorsford: We don't want to override the default grid column

app/views/new-record/delete-draft/confirm.html
```diff
@@ -0,0 +1,27 @@
...
+{% set formAction = "./../overview" | orReferrer(referrer) %}
```
edwardhorsford: No need for an action

edwardhorsford (changes requested):
Looking good.

The cancel link goes to the wrong place because of it using the wrong template.

app/views/_templates/_new-record.html
```diff
@@ -67,6 +67,9 @@
         {% if returnLink %}
           <p class="govuk-body"><a href="{{returnLink.href}}" class="govuk-link">{{returnLink.text}}</a></p>
         {% endif %}
+        {% if cancelLink %}
+          <p class="govuk-body"><a href="{{recordPath}}" class="govuk-link">Cancel and return to record</a></p>
+        {% endif %}
```
edwardhorsford: I don't think this is needed. The above `returnLink` handles it already.

In the view itself you can do:
`{% set returnLink = { href: 'foo', text: 'Cancel and return to record'} %}`

jesseyuen: You can? That's a bit fancy!

jesseyuen: IT WORKS!

edwardhorsford (approved):

PR #161, opened 2020-12-07
Content changes following ur

EmmaFrith:
**change 'get' to 'request' …**

- the expectations were not clearly set for one provider. they thought by logging in they'd get a list of TRN numbers. 
 
**change content and formatting in data requirements section …**

- change formatting (existing formatting was used to create different sections for different teaching routes - but we're finding that the data requirements for each route are more or less the same - table not really required
- remove phone number as no longer required
- use 'complete their record' rather than 'register them', because some data might not be needed to initially register someone
- remove 'given names' for consistency in other areas
- put UK and non-UK degrees in one section with 'if' options, as they're variables within the same requirement 
- use inset text to differentiate supporting guidance from data requirements list
- add a section about allocations, as this is needed for some routes 

**training record doesn't work …**

- use 'about their training' in task list and review pages, because it's confusing to use both 'training record' (currently in the task list) AND 'trainee records' (currently describes the list of trainees)
- change section heading 'training details' and relevant confirm/review content so it's clearer what this section is about
- use 'trainee progress' as a catch all for now when reviewing existing records

### DIFF
--- a/app/views/data-requirements.html
+++ b/app/views/data-requirements.html
@@ -129,7 +129,7 @@
 
     <h2 class='govuk-heading-m'>Allocations information</h2>
 
-    <p class="govuk-body">If the course has allocated places, you'll need to know whether the trainee is using an allocated place.</p>
+    <p class="govuk-body">If the course has allocated places, you’ll need to know whether the trainee is using an allocated place.</p>
 
   </div>
 </div>
--- a/app/views/data-requirements.html
+++ b/app/views/data-requirements.html
@@ -69,7 +69,7 @@
 
     <h2 class='govuk-heading-m'>Degrees</h2>
 
-    <h3 class='govuk-heading-s'>If it's a UK degree</h3>
+    <h3 class='govuk-heading-s'>If it’s a UK degree</h3>
 
     <ul class='govuk-list'>
       <li>
--- a/app/views/data-requirements.html
+++ b/app/views/data-requirements.html
@@ -10,7 +10,7 @@
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-l govuk-!-margin-bottom-9">{{pageHeading}}</h1>
 
-    <p class='govuk-body'>You'll need the following information about a trainee to complete their record.</p>
+    <p class='govuk-body'>You’ll need the following information about a trainee to complete their record.</p>
 
     <h2 class="govuk-heading-m">Personal details</h2>
 
--- a/app/views/data-requirements.html
+++ b/app/views/data-requirements.html
@@ -8,194 +8,129 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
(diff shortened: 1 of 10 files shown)


app/views/data-requirements.html
```diff
@@ -8,194 +8,129 @@
...
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-l">{{pageHeading}}</h1>
-    <ul class="govuk-list govuk-list--dash govuk-!-margin-bottom-8">
-      <li><a class="govuk-link" href="#section-all-routes">Standard data for all trainees</a></li>
-      <li><a class="govuk-link" href="#section-assessment-only">Specific data for assessment-only candidates</a></li>
-      <li><a class="govuk-link" href="#section-provider-led">Specific data for provider-led trainees</a></li>
+    <h1 class="govuk-heading-l govuk-!-margin-bottom-9">{{pageHeading}}</h1>
+
+    <p class='govuk-body'>You'll need the following information about a trainee to complete their record.</p>
```
edwardhorsford:
```suggestion
    <p class='govuk-body'>You’ll need the following information about a trainee to complete their record.</p>
```

app/views/data-requirements.html
```diff
@@ -8,194 +8,129 @@
...
+      <li>
+        Disability
+      </li>
+    </ul>
+
+    <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
+
+    <h2 class='govuk-heading-m'>Degrees</h2>
+
+    <h3 class='govuk-heading-s'>If it's a UK degree</h3>
```
edwardhorsford:
```suggestion
    <h3 class='govuk-heading-s'>If it’s a UK degree</h3>
```

app/views/data-requirements.html
```diff
@@ -8,194 +8,129 @@
...
+      </li>
+      <li>
+        Graduation year
+      </li>
+      <li>
+        Grade
+      </li>
+    </ul>
+
+    <h3 class='govuk-heading-s'>If it's not a UK degree</h3>
```
edwardhorsford:
```suggestion
    <h3 class='govuk-heading-s'>If it’s not a UK degree</h3>
```

app/views/data-requirements.html
```diff
@@ -8,194 +8,129 @@
...
+      <li>
+        End date
+      </li>
+    </ul>
+
+    <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
+
+    <h2 class='govuk-heading-m'>Allocations information</h2>
+
+    <p class="govuk-body">If the course has allocated places, you'll need to know whether the trainee is using an allocated place.</p>
```
edwardhorsford:
```suggestion
    <p class="govuk-body">If the course has allocated places, you’ll need to know whether the trainee is using an allocated place.</p>
```

edwardhorsford (changes requested):

edwardhorsford (approved):

PR #163, opened 2020-12-08
uppercase versus lowercase

EmmaFrith:
@johndoates pointed out to me the other day that we have to tendency to not utilise our service names as verbs, continually repeating the service name as a proper noun rather than a verb, resulting in a seemingly random capital letter mid sentence 

up for trying a different approach, so have used the service name as a verb in this heading

Screenshot 2020-12-08 at 13 05 54

### DIFF
--- a/app/views/sign-in.html
+++ b/app/views/sign-in.html
@@ -4,7 +4,7 @@
 {% set hideNav = true %}
 {% set hidePrimaryNav = true %}
 
-{% set pageHeading = 'Sign in to Register trainee teachers' %}
+{% set pageHeading = 'Sign in to register trainee teachers' %}
 
 {% block content %}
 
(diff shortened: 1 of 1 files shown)


edwardhorsford:
My initial reaction is that it looks a bit weird to me - but that might be because I'm used to seeing service names treated as proper nouns.

jesseyuen:
I suspect it's too literal a translation of the whole 'services should be verbs not nouns' thing to treat service names as verbs in most contexts. This looks weird.

johndoates:
Emma and I discussed this because the GOV.UK style guide does encourage this extension of the concept of ‘services as verbs’. It says this about service names:

> Lower case if the service name starts with a verb - write the sentence so the user knows what action they can take. For example: You can [visit someone in prison](https://www.gov.uk/prison-visits) by booking online.
> 
> Only use upper case if the name of the service you’re referring to contains a named thing. For example: [You can apply for Marriage Allowance](https://www.gov.uk/apply-marriage-allowance).

Although there's probably a discussion to be had about whether this approach will work in all contexts, I'd always take the style guide as the default and work from there.

In this case, I’d say that ‘Sign in to register trainee teachers’ is much better than ‘Sign in to Register trainee teachers’. The user is signing in so that they can register trainee teachers – so it makes sense as an action – plus you avoid the awkward capital letter in the middle of the H1.

I'd suggest following this approach through into the body text though. The use of 'DfE Sign In' is probably something to tackle another time, but I'd get rid of the capital on 'register'. The easiest way to do that would be just "Use DfE Sign In to access your account."

johndoates:
This is beyond the scope of this discussion, but it seems to me that this issue is only arising at all because of the way the journey works. 

It seems likely that the simpler wording ‘Register trainee teachers’ is avoided here because the previous page is called that. Is there any reason why /start-page can’t continue straight to the DfE sign in without this interim page?

adamsilver:
tagging in @duncanjbrown as he might know of why there is an interim page – I have a vague recollection that this was something to do with technical/security constraints

johndoates:
Just as a point of comparison, here's the equivalent page for signing in to manage applications. It was written before my time, but it does refer to the service as an action rather than as a thing: "You must sign in to your account to manage teacher training applications."

image

duncanjbrown:
I think the only constraint is that the sign in link be a button as the initial OAuth request must be a `POST` https://github.com/DFE-Digital/apply-for-teacher-training/pull/636

edwardhorsford:
There's a couple good reasons for it to be a separate page - it gives us something to link to when users sign out - that is within our control.

EmmaFrith:
> I'd suggest following this approach through into the body text though. The use of 'DfE Sign In' is probably something to tackle another time, but I'd get rid of the capital on 'register'. The easiest way to do that would be just "Use DfE Sign In to access your account."

I think I used "Register account" to differentiate this account from the "DfE Sign-in account". Maybe that's me overthinking it though. Thoughts @johndoates?

EmmaFrith:
> There's a couple good reasons for it to be a separate page - it gives us something to link to when users sign out - that is within our control.

Couldn't the start page also be somewhere for us to link to though @edwardhorsford?

PR #166, opened 2020-12-11
Initial support for letting users do bulk actions

edwardhorsford:
A common need in our service is for users to recommend a cohort of trainees for QTS. In many cases these trainees will have the same QTS date.

With our current service, to recommend a bunch of trainees, you'd have to go in to each one in turn and perform the same actions again and again - a highly repetitive and time consuming process.

This PR introduces initial support for bulk actions to make this easier. Users can pick an action, a set of filters, and then select from filtered trainees.

Initially we're supporting submitting for trn and recommending for QTS - the two most commonly requested actions. We could possibly consider more in the future.

I'm not entirely happy with how we link to this flow - but it's good enough for a first pass. Ultimately contextually relevant links may be better 'Recommend these trainees for QTS', etc.

I've also built internal support for _starting_ with a predefined list of trainees. So in the future we may be able to have other ui for picking trainees, and then jumping straight to doing an action on them.

### Bulk actions linked from header
Screenshot 2020-12-18 at 12 50 42

### Choosing an action
Screenshot 2020-12-18 at 12 50 49

### Filtering trainees
Screenshot 2020-12-18 at 12 50 57

### Selecting specific trainees
Screenshot 2020-12-18 at 12 51 05

### Collecting a date (QTS only)
Screenshot 2020-12-18 at 12 51 15

### Confirming data
Screenshot 2020-12-18 at 12 51 20

### Confirmation
Screenshot 2020-12-18 at 12 51 28

### DIFF
--- a/app/views/_includes/forms/bulk/date.html
+++ b/app/views/_includes/forms/bulk/date.html
@@ -47,9 +47,6 @@
       classes: "govuk-fieldset__legend--l govuk-!-margin-bottom-4"
     }
   },
-  hint: {
-      text: 'All selected trainees will be marked as having met the standards on this day. If your trainees met the standards on different days, go back and select trainees in batches.'
-    },
   items: [
     {
       text: "Today"
--- a/app/views/_includes/forms/bulk/date.html
+++ b/app/views/_includes/forms/bulk/date.html
@@ -38,7 +38,7 @@
   }) }}
 {% endset %}
 
-
+<span class="govuk-caption-l">{{data.bulk.action}}</span>
 {{ govukRadios({
   fieldset: {
     legend: {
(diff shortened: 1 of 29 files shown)


app/views/_includes/forms/bulk/date.html
```diff
@@ -0,0 +1,71 @@
...
+{{ govukRadios({
+  fieldset: {
+    legend: {
+      text: pageHeading,
+      isPageHeading: true,
+      classes: "govuk-fieldset__legend--l govuk-!-margin-bottom-4"
+    }
+  },
+  hint: {
+      text: 'All selected trainees will be marked as having met the standards on this day. If your trainees met the standards on different days, go back and select trainees in batches.'
```
edwardhorsford, line 51:
```suggestion
      text: 'All selected trainees will be marked as having met the standards on this day; If your trainees met the standards on different days, go back and select trainees in batches.'
```

PR #169, opened 2020-12-15
Hiding the failure options from the training outcome flow

jesseyuen:
This PR:

- Skips the question: "Have they met the standards for QTS?"
- Removes the outcome row from the summary card
- Updates to content in the check answers page

### DIFF
--- a/app/views/record.html
+++ b/app/views/record.html
@@ -25,7 +25,7 @@
         {% if record.status | getCanRecommendForQts %}
           {% if hasCommencementDate%}
             {{ govukButton({
-              text: "Record training outcome",
+              text: "Recommend trainee for QTS",
               href: referrer + '/qts/outcome-date',
               classes: "govuk-!-margin-bottom-0"
             }) }}
--- a/app/views/record.html
+++ b/app/views/record.html
@@ -26,7 +26,7 @@
           {% if hasCommencementDate%}
             {{ govukButton({
               text: "Record training outcome",
-              href: referrer + '/qts/outcome',
+              href: referrer + '/qts/outcome-date',
               classes: "govuk-!-margin-bottom-0"
             }) }}
           {% else %}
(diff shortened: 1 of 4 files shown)


app/views/record.html
```diff
@@ -26,7 +26,7 @@
           {% if hasCommencementDate%}
             {{ govukButton({
               text: "Record training outcome",
```
EmmaFrith: text: Recommend trainee for QTS

app/views/record/qts/outcome-date.html
```diff
@@ -2,13 +2,15 @@
...
+{# {% set pageHeading %}
   {% if data.record.qtsDetails.standardsAssessedOutcome == "Yes" %}
     When did they meet the standards? 
   {% else %} 
     When should they be recorded as having not met the standards?
   {% endif %}
-{% endset %}
+{% endset %} #}
+
+{% set pageHeading = 'When did they meet the standards?' %}
```
EmmaFrith: {% set pageHeading = 'When did they meet the QTS standards?' %}

app/views/record.html
EmmaFrith: I think the button should be more focussed on the action that you can now do

app/views/record/qts/outcome-date.html
EmmaFrith: To be clear what standards we're referring to

EmmaFrith: alternatively/ in addition - you could make the caption text more descriptive, i.e. 'Recommend Ameline Rodriguez for QTS'

edwardhorsford (changes requested):
The main CTA should probably be reverted to say 'Recommend for QTS' or similar.

edwardhorsford (approved):

app/views/record.html
jesseyuen: Good shout. Ta.

PR #171, opened 2020-12-16
Include trainee name with success message

edwardhorsford:
Done in colab with @EmmaFrith, include the trainee name in the success banner.


EmmaFrith (approved):
looks good!

PR #177, opened 2020-12-30
Add support for degree bursaries

edwardhorsford:
When trainees have more than 1 degree, providers need to select which degree is attracting the bursary.

This is done by adding a conditional degree selection page shown after adding a degree - but only if there's more than 1 degree already.

Most trainees won't have multiple degrees, so most will never see this section.

I'm a little unclear about what basis providers might pick between them, so the content will need further work once people are back from the Christmas break.

Screenshot 2021-01-07 at 16 44 10

Screenshot 2021-01-07 at 16 44 17


jesseyuen:
Perhaps showing my ignorance here... will there always be one that needs to be selected? For example, can a trainee have multiple degrees yet none be used for busary selection?

edwardhorsford:
> Perhaps showing my ignorance here... will there always be one that needs to be selected? For example, can a trainee have multiple degrees yet none be used for busary selection?

I've asked Ben, and we can raise at the next BAU meeting.

dankmitchell:
[Add degree selection for bursaries to prototype](https://trello.com/c/WiNEpzJQ/794-add-degree-selection-for-bursaries-to-prototype)

jesseyuen (approved):

PR #181, opened 2021-01-11
Support for users with multiple providers

edwardhorsford:
This PR adds more realistic support for providers in the prototype. In particular how we respond to users who belong to more than one provider - about 20% of our user base.

Key changes:
* Courses and records generated for a number of providers
* All records now relate to a provider
* The signed in user can belong to multiple providers
* When adding records, we assign a provider to that record

We're still deciding whether to adopt a 'blended' or 'hat' model for users who belong to multiple providers - this pr supports both in the ui so that we can try them out.

# Hat model

In this model (closest to DTTP and Publish), users must pick a provider to work as when signing in. To add trainees from different providers, they must back out and choose a different provider to work on - essentially putting on the 'hat' for the provider they're acting as.

## Choosing provider
Screenshot 2021-01-22 at 15 44 09

## Showing provider on draft record
Screenshot 2021-01-22 at 15 44 33

# Blended model

In this model, users who belong to multiple providers can see all records at once - but filter by a provider should they choose. This is probably the most flexible for ways of working, but with the downside that when adding a new record they'll need to choose a provider.

## Filtering by provider
Screenshot 2021-01-22 at 11 51 43

## Choosing a provider when adding a record
Screenshot 2021-01-22 at 12 00 54

## Showing provider on draft record
Screenshot 2021-01-22 at 12 01 17


dankmitchell:
[Prototype: support for multiple providers](https://trello.com/c/ytVA99y0/830-prototype-support-for-multiple-providers)

app/lib/utils.js
```diff
@@ -313,16 +357,22 @@ exports.filterRecords = (records, data, filters = {}) => {
...
+  filteredRecords = exports.filterBySignedIn(filteredRecords, data)
+
   // Only show records for training routes that are enabled
   let enabledTrainingRoutes = data.settings.enabledTrainingRoutes
 
   // Only show records for currently enabled routes or draft records
   filteredRecords = filteredRecords.filter(record => enabledTrainingRoutes.includes(record.route) || (record?.status === 'Draft'))
 
-  // Cycle not implimented yet
+  // Cycle not implimented yet§
```
jesseyuen: What's the `§` about?

app/views/_includes/trainee-record-card.njk
```diff
@@ -33,12 +33,18 @@
...
       </a>
     </h3>
+    
     {% if record.trainingDetails.traineeId %}
       <p class="govuk-caption-m govuk-!-font-size-16 app-application-card__id govuk-!-margin-bottom-0">Trainee ID: {{record.trainingDetails.traineeId}}</p>
     {% endif %}
     {% if record.trn %}
       <p class="govuk-caption-m govuk-!-font-size-16 app-application-card__id govuk-!-margin-bottom-0">TRN: {{record.trn}}</p>
     {% endif %}
+    {% if data.signedInProviders | length > 1 %}
```
jesseyuen, line 43: Wondering out loud if the provider name should be grouped with TRN and trainee ID in column 1 or or subject and route in column 2?

app/views/dfe-sign-in.html
```diff
@@ -0,0 +1,93 @@
...
+        },{
+          href: "/admin",
+          text: "Prototype admin"
+        }, {
+          href: "/admin/clear-data",
+          text: "Clear data"
+        }]
+      }
+    }) }}
+{% endblock %}
```
jesseyuen, line 93: Great to see this in the prototype! 👏

app/views/layout.html
```diff
@@ -78,16 +78,20 @@
...
-      </ul>
-
-    </nav>
+    {% if data.isHatModel %}
+      <nav class="moj-primary-navigation">
+        <ul class="moj-primary-navigation__list">
+          
+            <li class="moj-primary-navigation__item">
+              <a class="moj-primary-navigation__link" {{ 'aria-current="page"' | safe if navActive == 'provider' }} href="/select-provider">
+                {{ data.settings.userActiveProvider }}
```
jesseyuen, line 87: Wondering out loud if the provider name is obvious enough link text to say "this is where to click to swap hats". 

People _might_ assume the link would take them to more info about that provider. 🤔

app/views/layout.html
```diff
@@ -78,16 +78,20 @@
...
-        <li class="moj-primary-navigation__item">
-          <a class="moj-primary-navigation__link" {{ 'aria-current="page"' | safe if navActive == 'bulk' }} href="/bulk-action/new">
-            Bulk actions
-          </a>
-        </li>
-      </ul>
-
-    </nav>
+    {% if data.isHatModel %}
+      <nav class="moj-primary-navigation">
```
jesseyuen, line 82: Agree with the recent chatter on #bat_design, we could arguably just have one `<nav> `element in the header.

jesseyuen (approved):
Amazing work Ed ✨ 

Approving this with the "thinking out loud" comments for a chat at some point down the line.

PR #189, opened 2021-02-11
Add support for multiple skip links in header

edwardhorsford:
Our service has a number of filters you can apply to filter a group of records. We recently used the accessibility personas to review these pages - and felt it would be good to provide a way of skipping past the filters to the results.

In this pr I'm experimenting by adding a second skip link in to the header. This uses the _newish_ `:focus-within` property to show the skip-link div if either of the children links are focussed. This should gracefully degrade in browsers that don't support the property to just show two skip links in turn.

The css is mostly copied styles from the `govuk-skip-link` component, but applied to the header instead. If the browser supports `:focus-within` I then remove some of the default styles from the skip links themselves.

## Video:
dual-skip-links

## Without support for the property:
double-skip-link

### DIFF
--- a/app/assets/sass/overrides/_govuk-skip-link.scss
+++ b/app/assets/sass/overrides/_govuk-skip-link.scss
@@ -1,0 +1,101 @@
+$important: true;
+
+// This is basically a copy of the Design System govuk-visually-hidden-focusable mixin
+// except it targets the :focus-within pseudo selector rather than :focus
+@mixin visually-hidden-focus-withinable {
+  position: absolute if($important, !important, null);
+
+  width: 1px if($important, !important, null);
+  height: 1px if($important, !important, null);
+  // If margin is set to a negative value it can cause text to be announced in
+  // the wrong order in VoiceOver for OSX
+  margin: 0 if($important, !important, null);
+
+  overflow: hidden if($important, !important, null);
+  clip: rect(0 0 0 0) if($important, !important, null);
+  clip-path: inset(50%) if($important, !important, null);
+
+  // For long content, line feeds are not interpreted as spaces and small width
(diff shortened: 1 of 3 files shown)


app/views/records.html
```diff
@@ -4,6 +4,24 @@
 {% set backLink = 'false' %}
 {% set navActive = "records" %}
 
+{% block skipLink %}
+ 
+<div class="app-skip-link--container">
```
jesseyuen: Consider indenting `app-skip-link--container`

app/views/records.html
```diff
@@ -4,6 +4,24 @@
...
+  <span class="app-skip-link--item">
+    {{ govukSkipLink({
+      href: '#main-content',
+      text: 'Skip to main content'
+    }) }}
+  </span>
+  <span class="app-skip-link--item">
+    {{ govukSkipLink({
+      href: '#records-list',
+      text: 'Skip to results' if selectedFilters else 'Skip to records'
```
jesseyuen: Nice!

app/assets/sass/overrides/_govuk-skip-link.scss
```diff
@@ -0,0 +1,99 @@
+$important: true;
+
+// This is basically a copy of the Design System govuk-visually-hidden-focusable mixin
```
jesseyuen, line 3: Not related to your code but I find a bunch of the design system css so hard to read. 🧐

jesseyuen (approved):
This looks great Ed. I gave it a test in a legitimate ie11 instance too and the fallback works well.

app/views/records.html
edwardhorsford: Thanks! fixed.

PR #190, opened 2021-02-17
Make classes use correct BEM syntax

edwardhorsford:


app/assets/sass/overrides/_govuk-skip-link.scss
```diff
@@ -55,7 +55,7 @@ $important: true;
 // applies to browsers that support the :focus-within pseudo property. 
 // Using this instead of @supports as it doens’t support pseudo properties well yet.
 // Many of the styles within are from the .govuk-skip-link class
-.app-skip-link--container:focus-within, .app-skip-link--container:not(:focus-within) {
```
jesseyuen, line 58: In BEM-ese this `.app-skip-link__container` class suggests it has is a parent, block element called `.app-skip-link`

jesseyuen: ...but it doesn't.

jesseyuen (approved):

PR #195, opened 2021-02-24
Support extra qualifications

edwardhorsford:
This pr:
* Adds initial support for alternate qualifications per route
* Adds some filters to return the qualification for a given record
* Updates the  names of different routes to make them more consistent
* Enables 1 early years route and the school direct routes to start work on them
* Cleans up the training route data a bit
* Updates the generators to more heavily use the training route data


jesseyuen (approved):
A bit over my head but I like your intention so am happy to approve!

PR #197, opened 2021-02-25
Fix unsafe deletes and add ability to directly import json

edwardhorsford:
This fixes a bug where we were unsafely using `delete` to attempt to delete properties of objects that didn't exist.

I had a search through the codebase and added optional chaining to all our nested deletes.

I also added a hacky feature I used for testing it. Visiting the `direct-add` route will load in a json blob for the record. This was particularly helpful for manually testing - without having to generate new seed data.

### DIFF
--- a/app/data/direct-add-record.json
+++ b/app/data/direct-add-record.json
@@ -1,0 +1,100 @@
+{
+  "status": "Draft",
+  "events": {
+    "items": []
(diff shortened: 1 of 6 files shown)


jesseyuen (approved):
5 stars, would recommend Ed! He's a top sort.

PR #201, opened 2021-03-04
Add target to open link in new tab

jesseyuen:


edwardhorsford (changes requested):
I think this should probably mention in the link text that it's [going to open in a new tab](https://design-system.service.gov.uk/styles/typography/#opening-links-in-a-new-tab).

edwardhorsford (approved):

PR #202, opened 2021-03-05
Rename programme details to course details

edwardhorsford:
This renames programme details to course details across the prototype.

This is something we've wanted to do for a while, but touches a lot fo the app.

Reasons for change:
* Allows us to more closely match other language such as 'publish course'
* Hopefully has more distinction between these details (which are generic) and details that are specific to the trainee's training (like their start date).

I mostly did this with lots of find and replace. 

Suggest having a play with the review app.


app/views/_includes/summary-cards/course-details/publish-course.html
```diff
@@ -39,7 +39,7 @@
     actions: {
       items: [
         {
-          href: recordPath + "/programme-details" | addReferrer(referrer),
+          href: recordPath + "/course-details" | addReferrer(referrer),
           text: "Change course",
```
jesseyuen: Duplicate "course" here as it's also in the `visuallyHiddenText: "course"`

jesseyuen (approved):
Approved but one change requested.

PR #204, opened 2021-03-08
Add Apply data to seeds

edwardhorsford:
This updates the seeds to also generate some fake Apply records.

I've added a few extra fields that I think we'll get from Apply.


app/data/training-route-data.js
```diff
@@ -1,6 +1,12 @@
 // A non-exhaustive list of routes
 // Publish and non publish can overlap
 
+let applyRoutes = [
+  'Provider-led (postgrad)',
+  'School direct (salaried)',
+  'School direct (tuition fee)',
```
jesseyuen, line 7: Should there be a trailing comma here? `'School direct (tuition fee)',`

app/views/_includes/trainee-record-card.njk
```diff
@@ -28,7 +28,7 @@
           {% if record.status == "Draft" %}
             Draft record
           {% else %}
-            Name not known {# Shouldn't be possible - non drafts should always have a name #}
+          Name not known {# Shouldn't be possible - non drafts should always have a name #}
```
jesseyuen: Could/Should be indented again.

jesseyuen (approved):

app/data/training-route-data.js
edwardhorsford: It's not a problem, but will remove for consistency.

PR #223, opened 2021-03-26
[SPIKE] Schools autocomplete with large number of schools

AbigailMcP:
## What?

https://trello.com/c/7OGaCU6j/1327-spike-1day-autocomplete-for-schools-lookups

This PR is the output for the above spike - experimenting with providing a large list of schools to an autocomplete for UR.

## How?

**1. Switching to a standalone 'autocomplete' text input**
- The Accessible Autocomplete can be used either to progressively enhance a select input, or as a standalone text input. Mostly we use it as an enhancement, since this makes the non-JS fallback easy (it's just a select). However, this is not feasible with ~40k possible results for two (or more) reasons:
  - The non-JS fallback of a select with ~40k records is... no.
  - The autocomplete performance searching across this number of records is not good enough

**2. Doing the filtering ourselves**
- The autocomplete normally takes a simple array of strings as it's `source`. However, you can also provide it with a function. This function is passed two arguments:
  - `query` (string) - i.e. the search term from the user
  - `populateResults` (function) - i.e. a callback which expects to be passed an array of results to render
- This way, the filtering and sorting of results is completely down to the developer to handle (either on the server or client-side) and we can improve the performance.

**3. Actually improving the performance**
- Without a database, this is a little tricky. In order to achieve a performant text search, we need some kind of index.
- Enter https://lunrjs.com/ - a package for creating simple search indexes on documents (json)
- Creating the search index every time we need it would be slow, so here I'm generating it once when the app loads and saving it.
- Then the autocomplete can use this pre-built search index to quickly return matches

**4. Hooking it all up**
- The schools autocomplete makes a request to our server for the search index when the page loads
- When users enter a search term we call a (debounced) function which queries this index for our search term
- When the result is selected, a hidden field is populated with the UUID, and the name is shown in the autocomplete input
- The UUID is sent to the server to match the correct school (there are multiple schools with the same name so we can't just submit the name)

## And now?
- [ ] This doesn't solve the non-JS issue, this should be thought about in a separate spike. Make a ticket.
- [ ] This doesn't use the _actual_ schools list. Generate the correct json and include.
- [x] The default text 'No results found' is misleading on occasions when we're actually still searching. Make it say 'Loading...' or something


AbigailMcP:
[Spike [1day] autocomplete for schools lookups](https://trello.com/c/7OGaCU6j/1327-spike-1day-autocomplete-for-schools-lookups)

edwardhorsford (approved):
This is fantastic, thanks @AbigailMcP! Should be great to test with.

PR #226, opened 2021-03-30
Offer deferred date when withdrawing if trainee is deferred

edwardhorsford:
We've heard in user research that trainees will defer for some time, and then may decide to withdraw from the course. In such a situation the provider should use the original deferral date as the withdrawal date - as this is the last time they were on the course.

On our current withdrawal page, the date of deferral isn't visible - but is the likely one. This 'defers' to the deferral date if the trainee is currently deferred.

Screenshot 2021-04-07 at 17 43 37

Screenshot 2021-03-31 at 12 32 51

### DIFF
--- a/app/lib/utils.js
+++ b/app/lib/utils.js
@@ -180,6 +180,18 @@ exports.routeHasPublishCourses = function(record){
 // Records
 // -------------------------------------------------------------------
 
+exports.isAwarded = record =>{
+  return record.status == "QTS awarded" || record.status == "EYTS awarded"
+}
+
+exports.isDeferred = record =>{
+  return record.status == "Deferred"
+}
+
+exports.isWithdrawn = record =>{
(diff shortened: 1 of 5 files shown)


jesseyuen (approved):
I like this one.

edwardhorsford:
[Prototype: use deferral date when withdrawing a deferred trainee](https://trello.com/c/62O5bGT2/1386-prototype-use-deferral-date-when-withdrawing-a-deferred-trainee)

edwardhorsford:
@jesseyuen I've updated this after discussion with Shaun. @gregknight1 still needs to look at it too.

gregknight1:
@edwardhorsford - Think we can reduce the content on this:

We'll use the trainee's deferral date of 28 July 2020 as their withdrawal date.

To use a different date, [reinstate this trainee]() then withdraw them.

PR #231, opened 2021-03-31
Add hint to course selection page

edwardhorsford:
We've seen some users unsure where these courses are from, or why they're just *some* of the courses. This should help reinforce that these are from Publish and that they're for the route the user has selected.

Screenshot 2021-03-31 at 16 43 54

Note the 'stored' route name includes brackets around `postgrad` and `undergrad`. This looks a bit weird in a sentence so I've stripped them here.

### DIFF
--- a/app/views/_includes/forms/course-details/pick-course.html
+++ b/app/views/_includes/forms/course-details/pick-course.html
@@ -36,6 +36,10 @@ defer to an id if one exists. #}
     checked: checked(selectedCourse, "Other")
   }) %}
 
+  {% set hintText %}
+    Your {{record.route | lower}} courses in the Publish service
+  {% endset %}
+
   {{ govukRadios({
     fieldset: {
       legend: {
@@ -43,6 +47,9 @@ defer to an id if one exists. #}
         isPageHeading: true,
         classes: "govuk-fieldset__legend--l"
       }
+    },
+    hint: {
+      text: hintText
     },
     items: courseItems
   } | decorateAttributes(record, "record.selectedCourseTemp")) }}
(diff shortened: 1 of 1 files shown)


edwardhorsford:
@gregknight1 can we take a look at this?

gregknight1:
@edwardhorsford The hint text is fine. I do have reservations about using 'postgrad' and 'undergrad', though. Think we should be using postgraduate and undergraduate - unless we have evidence that all of our users use the abbreviated versions.

Realise that we may have to make that change elsewhere (not sure how often they appear abbreviated) so I'm happy to raise that as a separate card.

edwardhorsford:
> @edwardhorsford The hint text is fine. I do have reservations about using 'postgrad' and 'undergrad', though. Think we should be using postgraduate and undergraduate - unless we have evidence that all of our users use the abbreviated versions.
> 
> Realise that we may have to make that change elsewhere (not sure how often they appear abbreviated) so I'm happy to raise that as a separate card.

Let's discuss separately from this. The main reason I used the abbreviation was because of how long the route names become - which particularly impacts the sort filters.

gregknight1:
> > @edwardhorsford The hint text is fine. I do have reservations about using 'postgrad' and 'undergrad', though. Think we should be using postgraduate and undergraduate - unless we have evidence that all of our users use the abbreviated versions.
> > Realise that we may have to make that change elsewhere (not sure how often they appear abbreviated) so I'm happy to raise that as a separate card.
> 
> Let's discuss separately from this. The main reason I used the abbreviation was because of how long the route names become - which particularly impacts the sort filters.

Yes, that makes sense to look at it separately.

PR #212, opened 2021-03-17
Add school direct routes

edwardhorsford:
This PR adds two new training routes to the prototype.

The main new thing for these routes is that both collect a lead school, and one collects an employing school.

As part of this I've added a schools generator - though we might end up replacing this with real schools data.

We'll need to do further work to see if an autocomplete is viable for searching from the ~30k school records - and design a no-js fallback.

As part of a future pr we can use the same autocomplete I've made here for the placement schools

Things not done as part of this pr:
* placement schools lookup
* full search across ~30k schools
* We may come back in a future pr and make schools an entirely separate section in the task list

## Screenshots 

Lead school selection
Screenshot 2021-03-23 at 15 15 28

Lead school selection when searching
Screenshot 2021-03-23 at 15 15 35

Employing school selection
Screenshot 2021-03-23 at 15 15 44

Summary card for training details
Screenshot 2021-03-23 at 15 15 56

Schools on a trainee record
Screenshot 2021-03-23 at 15 16 13

PR #237, opened 2021-04-07
No js school search WIP

edwardhorsford:
The first school search (#224 #212) just had an autocomplete for picking schools. But we also need a no-js base to progressively enhance from.

In this iteration, the default input acts as a search. If a user doesn't have js or if they type a query but don't pick an item from the autocomplete, then we return them to a results page which lists a truncated set of schools to pick from, with an option to change the search.

Search input:
Screenshot 2021-04-07 at 17 15 01

Results:
Screenshot 2021-04-07 at 17 15 21

There's lots of gotchas in how the prototype handles edge cases and some unresolved things:
* Searching again is done as a conditional input - but for no-js this would be visible. It would be possible for a user to pick a school, change their mind and attempt to search (without changing the radio to 'search again') - this would end up selecting the school they first picked.
* Edit journey feels complicated. I think we could start by *not* prepopulating when editing - treat each time through this flow as a *new* search.

Videos:
school-select-no-js-fallback

Changing your search:
school-select-no-js-fallback-change-search

As part of this I also tweaked the Lunr search to:
* strip punctuation from queries
* Bias towards the last search term being fully present
* Prune low scoring search results


jesseyuen (approved):
Quite a lot going on here that I dont quite get but I'm ok with that. Have had a play it all looks good man. Nice work on a tricky flow man.

PR #241, opened 2021-04-20
A first pass on the 'requires attention' section WIP

jesseyuen:
This PR:
- adds an early concept for how might might show alerts that require our users attention. I expect this design to change. I mean CIRCLES! 
- adds a feature flag in the /admin section

We'll need @gregknight1 to input from a content perspective.

homapge

### DIFF
--- a/app/views/home.html
+++ b/app/views/home.html
@@ -54,7 +54,7 @@
             <div class="app-action-required-item">
               <span class="app-action-required-item__count">{{applyDraftCount}}</span>
               <div class="app-action-required-item__content">
-                <a href="/records?filterStatus=Draft" class="govuk-link app-action-required-item__text">new Apply drafts are ready to review</a>
+                <a href="/records?filterStatus=Draft" class="govuk-link app-action-required-item__text">Apply drafts ready for review</a>
                 <span class="app-action-required-item__hint-text govuk-hint govuk-!-font-size-16">These are candidates who have recenlty applied to a course</span>
               </div>
             </div>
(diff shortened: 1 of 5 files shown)


app/views/home.html
```diff
@@ -42,6 +45,30 @@ <h2 class="govuk-heading-m"><a href="/data-requirements" class="govuk-link govuk
 
     <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
 
+    {% if data.settings.showRequiresAttentionSection %}
```
edwardhorsford: This should also check if the counts of either of your two variables are greater than 0

app/views/home.html
```diff
@@ -42,6 +45,30 @@ <h2 class="govuk-heading-m"><a href="/data-requirements" class="govuk-link govuk
...
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds-from-desktop">
+          <h2 class="govuk-heading-m">Requires your attention</h2>
+          <div class="app-attention-link">
+            <span class="app-attention-link__count">{{applyDraftCount}}</span>
+            <div class="app-attention-link__content">
+              <a href="/records?filterStatus=Draft" class="govuk-link app-attention-link__text govuk-!-font-size-24">Apply drafts are ready to review</a>
```
edwardhorsford: Do you need the manual font size here? could `app-attention-link__text` set the size?

edwardhorsford (commented):
Looking good man!

I'm not sure about the name `attention-link`.

app/views/home.html
jesseyuen: And I imagine also a check within to only show each link when the count is > 0

edwardhorsford (approved):

gregknight1:
Sorry I'm late to this @jesseyuen! Few observations and questions:

- Think 'attention' is a bit vague - maybe the heading could be something like 'Action required' as they need to do something. Do we say anything similar anywhere else in the service?
- Apply drafts link could be: (7) Apply drafts ready for review
- cycle end link: (21) Records for review before cycle ends

Happy to chat about any of this. Looking good.

jesseyuen:
> Sorry I'm late to this @jesseyuen! Few observations and questions:
> 
> * Think 'attention' is a bit vague - maybe the heading could be something like 'Action required' as they need to do something. Do we say anything similar anywhere else in the service?
> * Apply drafts link could be: (7) Apply drafts ready for review
> * cycle end link: (21) Records for review before cycle ends
> 
> Happy to chat about any of this. Looking good.

Thanks @gregknight1. Will add those changes in.

PR #243, opened 2021-04-23
Reorder degree fields and split in to two templates

edwardhorsford:
Changes:
* Give degree details pages specific headings if they're UK or Non-UK
* Reorder degree fields
* Reorder rows on summary lists to match
* Refactor templates

---

We've seen users frequently attempt to include `hons` as part of the degree type - but this is captured separately as part of the grade.

One hypothesis for helping with this is to move the `degree grade` field next to the `degree type` field. The order is otherwise mostly arbitrary / what looked 'right' so it's no real problem to move it.

I've also taken the opportunity to clean up the templates a little - splitting the template in to one for uk and one for non-uk.

Screenshot 2021-04-23 at 12 48 29

UK degree details form
Screenshot 2021-04-23 at 12 53 39

Non-UK degree details form
Screenshot 2021-04-23 at 12 53 22


app/views/_includes/forms/degree/details-international.html
```diff
@@ -0,0 +1,93 @@
...
+      classes: "govuk-fieldset__legend--s"
+    }
+  },
+  hint: {
+    text: "Formerly UK NARIC"
+  },
+  items: enicItems
+} | decorateAttributes(degreeTemp, "degreeTemp.typeInt")) }}
+
+{# Graduation year year #}
```
jesseyuen:
```suggestion
{# Graduation year #}
```

jesseyuen (approved):

PR #244, opened 2021-04-23
Cleanup

edwardhorsford:
A mixed collection of bug fixes.

Including:
* Homepage now using a filtered set of records which should match what's shown on the records list
* Including counting `Apply drafts` in the `drafts` status card
* Using `orReferrer()` instead of `referrer or` as we have more complex logic we need to test
* Adding a link to a newly created record
* Filtering by draft trainees now includes `Apply draft` records

### DIFF
--- a/app/views/new-record/check-record-apply-grouped-sections.html
+++ b/app/views/new-record/check-record-apply-grouped-sections.html
@@ -4,7 +4,7 @@
 {% set formAction = "./save" | addReferrer(referrer) %}
 {% set hideReturnLink = true %}
 {% set showIncomplete = true %}
-{% set backLink = referrer or './overview' %}
+{% set backLink = './overview' | orReferer(referrer) %}
 {% set backText = 'Back to draft record' %}
 {% set gridColumn = 'govuk-grid-column-full' %}
 
--- a/app/views/new-record/check-record-apply-grouped-sections.html
+++ b/app/views/new-record/check-record-apply-grouped-sections.html
@@ -4,7 +4,7 @@
 {% set formAction = "./save" | addReferrer(referrer) %}
 {% set hideReturnLink = true %}
 {% set showIncomplete = true %}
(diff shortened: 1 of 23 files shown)


app/views/new-record/check-record-apply-grouped-sections.html
```diff
@@ -4,7 +4,7 @@
 {% set formAction = "./save" | addReferrer(referrer) %}
 {% set hideReturnLink = true %}
 {% set showIncomplete = true %}
-{% set backLink = referrer or './overview' %}
+{% set backLink = './overview' | orReferrer(referrer) %}
```
jesseyuen, line 7: What's the benefit of swapping these around bud? How is `orReferrer(referrer)` different?

app/views/new-record/submitted.html
```diff
@@ -23,7 +23,13 @@
 
 <p class="govuk-body">The Department for Education (DfE) will issue a teacher reference number (TRN) within 3 working days. </p>
 
-<p class="govuk-body govuk-!-margin-bottom-7">DfE will contact trainees to give them their TRN.</p>
+<p class="govuk-body">DfE will contact trainees to give them their TRN.</p>
+
+{% if data.submittedRecordId %}
+  <p class="govuk-body govuk-!-margin-bottom-7">
+    <a class="govuk-link" href="/record/{{data.submittedRecordId}}">View {{record.personalDetails.shortName}}’s record</a>
```
jesseyuen, line 30: 👏

jesseyuen (approved):

app/views/new-record/check-record-apply-grouped-sections.html
edwardhorsford: The filter `orReferrer` is now doing more than just swapping one url for another. I added support for chains of referring urls and this filter now gets the _last_ item for the chain.

With all that said `orReferrer` is a poor name, as the referrer is given priority. Suggestions welcome for a better name.

PR #245, opened 2021-04-26
Move schools in to their own section

edwardhorsford:
This makes a new section 'Schools' for collecting lead and employing school.

Why do this?
When we first added schools I debated with myself whether it should be a new section or just part of training details.

In favour of it being within training details is that it avoids another very short section (increasing time needed to complete a record). The downside of this is it's less _obvious_ that training details contains schools. Whether schools are asked for or not depends on the route chosen (school direct only at the moment) so there's also something about the task list unambiguously showing the sections that relate _to the route selected_. Our developers are starting off with building it as a separate section (as this is easier) so this now makes the prototype match what's planned to be built.

Screenshot 2021-04-26 at 16 55 11
Screenshot 2021-04-26 at 16 55 28


jesseyuen (approved):
LGTM :)

PR #246, opened 2021-04-29
Improve page load times

AbigailMcP:
Just testing some things out!

### DIFF
--- a/app/routes.js
+++ b/app/routes.js
@@ -7,7 +7,6 @@ const router = express.Router()
 const url = require('url')
 const utils = require('./lib/utils')
 const fs = require('fs')
-const searchIndex = require('./lib/search-index.json')
 
 // =============================================================================
 // Catch all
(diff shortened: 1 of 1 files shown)


AbigailMcP:
@edwardhorsford it's still quite slow, but:

Loading trainee index down from 522ms -> 174ms
Loading an individual trainee down from 318ms -> 81.5ms

edwardhorsford (approved):

edwardhorsford:
@AbigailMcP thanks so much for looking at this!

edwardhorsford:
Do you think there's a similar thing to do for the `shared-record-routes.js` where we read in the schools data?

AbigailMcP:
> Do you think there's a similar thing to do for the `shared-record-routes.js` where we read in the schools data?

Short answer: yes!

Long answer: yes, but I just attempted a refactor and broke everything 😂 - I'm going to merge this in as an incremental improvement and park the second bit for now.

PR #247, opened 2021-04-30
Add support for multiple subjects

edwardhorsford:
We need to support trainees having up to 3 subject specialisms. Most trainees will have only one, but some common multiple subject courses are modern languages and sciences.

At the moment I'm concatting the subjects in the form `A with B` or `A with B and with C`. This isn't super ideal, but we could look to address in the future if it's a problem. In DTTP subject A would generate an `allocation subject` - perhaps we could display these instead.

Core changes:
* Update seeds to output 1 - 3 subjects.
* Publish / Apply courses only have up to 2 subjects.
* Bias multiple subjects to common combinations
* Collect multiple subject using details and new autocompletes
* Dynamically combine the subject names
* * concat with `with`
* * lowercase second subject (except languages)
* * shorten some subjects for readability
* When displaying in summary lists, pluralise the key if there's more than one subject.

register-multiple-subjects2

Screenshot 2021-04-30 at 12 08 04

Screenshot 2021-04-30 at 12 08 14

Screenshot 2021-04-30 at 12 10 28


edwardhorsford:
[Add support for teachers training in multiple subjects](https://trello.com/c/1vCmBtEP/1618-add-support-for-teachers-training-in-multiple-subjects)

app/routes/shared-edit-routes.js
```diff
@@ -111,8 +111,6 @@ module.exports = router => {
     let recordPath = utils.getRecordPath(req)
     let referrer = utils.getReferrer(req.query.referrer)
 
-    console.log('hello world')
```
jesseyuen, line 114: Goodbye `console.log`

jesseyuen (approved):
Looks good Ed! I say merge it in

PR #261, opened 2021-05-10
Simplify labels for age ranges

edwardhorsford:
After:
Screenshot 2021-05-10 at 16 33 37

Before:
Screenshot 2021-05-10 at 16 39 19


jesseyuen (approved):
This is good!

PR #271, opened 2021-05-21
Make new guidance page and fold data requirements in to it

edwardhorsford:
Replace our existing `Check what data you need` page with a more general `Guidance` page which includes that content and some more general content about Register.


gregknight1:
Looks good @edwardhorsford - If you merge this, I can make changes directly.

PR #283, opened 2021-06-09
Update bursary.html

gregknight1:
Make the 2 sentences of content into 1 para and include a link to DfE funding manual.
Include scholarship in the No radio.

Also wanted to change the content on the confirm page to 'Not bursary funded'

### DIFF
--- a/app/views/_includes/forms/funding/bursary.html
+++ b/app/views/_includes/forms/funding/bursary.html
@@ -5,7 +5,7 @@
   The course you have selected has a bursary available of {{record | getBursaryValue | currency }} for {{ bursary.subject | dynamicLowercase }}. You need to check if the trainee is eligible for this bursary.
 </p>
 
-<p class="govuk-body"><a href="https://www.gov.uk/government/publications/initial-teacher-training-itt-bursary-funding-manual/initial-teacher-training-bursaries-funding-manual-2021-to-2022-academic-year#training-bursary-award-and-eligibility" class="govuk-link">DfE bursary guidance for academic year 2021 to 2022</a></p>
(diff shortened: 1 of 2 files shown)


app/views/_includes/forms/funding/bursary.html
```diff
@@ -2,12 +2,10 @@ <h1 class="govuk-heading-l">{{pageHeading}}</h1>
 {% set bursary = record | getBursary %}
 
 <p class="govuk-body">
-  The course you have selected has a bursary available of {{record | getBursaryValue | currency }} for {{ bursary.subject | dynamicLowercase }}.
+  The course you have selected has a bursary available of {{record | getBursaryValue | currency }} for {{ bursary.subject | dynamicLowercase }}.You need to check if the trainee is eligible for this bursary.
```
edwardhorsford:
```suggestion
  The course you have selected has a bursary available of {{record | getBursaryValue | currency }} for {{ bursary.subject | dynamicLowercase }}. You need to check if the trainee is eligible for this bursary.
```

app/views/_includes/forms/funding/bursary.html
```diff
@@ -28,7 +26,7 @@ <h1 class="govuk-heading-l">{{pageHeading}}</h1>
       value: "false"
     },
     {
-      text: "No, the trainee is self funded or not eligible",
+      text: "No, the trainee is self funded, not eligible or has a scholarship",
```
edwardhorsford, line 28: I wonder about making this more generic `No, not applying`, etc - and then considering hint text with some possible reasons.

edwardhorsford (changes requested):

app/views/_includes/forms/funding/bursary.html
gregknight1: Yes maybe - I originally had "No, the trainee is self funded, not eligible or is apply for a scholarship", but it was too long. Also considered bullets but that looked weird. 

Let's chat at 2, I also haven't figured out how to amend the dynamic content on confirmation pages, so would be good to learn that too. @edwardhorsford

edwardhorsford (approved):

PR #294, opened 2021-06-25
Update start date question to use course start date if available

edwardhorsford:
This updates our trainee start date question to do two things:
* Offer to use the course start date if available
* Use tenseless text to avoid implying a trainee needs to have started a course to be registered

We can only offer to use the course start date if we have it available. Right now this section comes after the course details section, but we don't requite course details to be done first - and I'm keen to avoid more conditionality.

If a course start date is available we'll ask like this:
trainee-start-date-question-2

If a course start date isn't available, we'll just show the date input, but with similar text as used by the radio question:
Screenshot 2021-06-25 at 16 11 22

Have also updated the record edit page for start date to match:
Screenshot 2021-06-25 at 16 11 42

### DIFF
--- a/app/routes/shared-edit-routes.js
+++ b/app/routes/shared-edit-routes.js
@@ -333,11 +333,8 @@ module.exports = router => {
     let record = data.record
     let recordPath = utils.getRecordPath(req)
     let referrer = utils.getReferrer(req.query.referrer)
-
     let enabledRoutes = data.settings.enabledTrainingRoutes
     let route = record?.route
-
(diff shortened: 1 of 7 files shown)


gregknight1:
@edwardhorsford Looking at the hint text again - I'd like to change it to:

_**This is the date the trainee starts on the course. This can be different from the course start date.**_

Think what we've got implies that it's always a different date, which could cause confusion.

gregknight1 (approved):
Looks good - thanks.

PR #308, opened 2021-07-20
updates homepage layout

ralph-hawkins:
### v1

Changes page heading from the service name (‘Register trainee teachers’) to ’Your trainee teachers’. This is to
- reduce confusion with the start page that has the same heading
- make the purpose of the page a little clearer

Adds the organisation name to the heading, this sets context for the 'Your' in 'Your trainee teachers' and will for users that belong to more than one organisation. Arguably it doesn't add a lot.

Makes the heading XL for consistency with the rest of the service and gives room in heading hierarchy for sections in the page.

Updates the nav to refer to 'trainees' consistently instead of 'records'. The service currently uses 'records' and 'trainees' either inconsistently or in a way thats pretty subtle — I personally find 'record' to be too general and find it helpful to use 'trainee'.

Removes the lead-in paragraph, this doesn't need to be seen repeatedly by users.

Styles 'Add a trainee' as a link and puts it into the new 'Draft trainees' section.

Adds sections:
- Draft trainees
- Registered trainees
- Guidance

This gives a stronger distinction between 'draft' and non-draft trainee records. This is consistent with the top nav and helps with prioritise incomplete records as a group.

The three actions in draft are shown as a simple list in part because they are conditional on whether the user has any draft trainees or records imported from Apply. If the user hasn't created any draft trainees but has records from Apply we'd only show them the 'See draft trainees imported from Apply' link because the 'all draft' link would do the exact same thing. This will also make prioritisation clearer.

Removes the middle section of the page with 3 links as the relationship between them wasn't clear and the new sections give a clearer home for the 'Registered trainees' and 'Guidance' links.

Removes the feedback link as we have it at the top of the page.

Adds a link to the current years trainees.

Reorders the status cards. Given that there are now 6 it felt odd for 'Qualification recommended' and 'Qualification awarded' to be in different rows.

Made the status cards use css-grid so they adapt to different viewport widths more nicely.

Splits the guidance links. This is so the guidance title are more useful than 'guidance'. For now the two links go to different sections in the same page, but the guidance page should be split.

Things to consider
- is the hierarchy correct, should registered trainees sit above  draft?
- we need to update the welcome email
- we need to split the guidance

localhost_3000_home(screenshot for design history — short) (2)

### v2

After speaking to the team, I've made a few changes:

Changed 'Add a trainee' to 'Create a trainee record'.  There's concern the link isn't strong enough as a call to action. I've tried to make the wording clearer, but not addressed the visual strength of this as its not clear when this is the primary action above modifying existing drafts.

Reverted the order of the statues. This means the 'Deferred' and 'Withdrawn' status-cards need to be their own row. I've made the grid 4 x 2 (2 x 3 at mobile width). This is a little less visually tidy, but probably makes the flow from pending to qualification awarded clearer and doesn't give overdue prominence to 'Withdrawn' and 'Deferred' statuses.

Changed 'Check what data you need' to 'Check what data you need to provide', this is a clearer title.

Desktop:

localhost_3000_home(screenshot for design history — short) (3)

Mobile example:

localhost_3000_home(iPhone 6_7_8)


app/views/home.html
```diff
@@ -1,156 +1,84 @@
 {% extends "layout.html" %}
 
 {% set navActive = "home" %}
+{% set totalDraftCount = data.filteredRecords | where("status", ["Draft", "Apply draft"] ) | length %}
 {% set applyDraftCount = data.filteredRecords | where('status', 'Apply draft') | length %}
 {% set recordsToBeReviewedCount = 0 %}
```
edwardhorsford: Can this be deleted?

app/views/home.html
```diff
@@ -1,156 +1,84 @@
...
-        }) }}
-      </div>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds-from-desktop">
+      <span class="govuk-caption-l">{{ data.signedInProviders | andSeparate }}</span>
+      <h1 class="govuk-heading-xl">{{ pageHeading }}</h1>
+      <h2 class="govuk-heading-l">Draft trainees</h2>
+      <ul class="govuk-list govuk-list--spaced govuk-!-margin-bottom-0">
+        {% if totalDraftCount != applyDraftCount %}
+          <li><a href="/drafts">See all draft trainees ({{ totalDraftCount }})</a></li>
```
edwardhorsford:
```suggestion
          <li><a class="govuk-link" href="/drafts">See all draft trainees ({{ totalDraftCount }})</a></li>
```

app/views/home.html
```diff
@@ -1,156 +1,84 @@
...
+        {%  endif %}
+        {% if applyDraftCount > 0 %}
+          <li><a href="/drafts?filterSource=Apply">See draft trainees imported from Apply ({{applyDraftCount}})</a></li>
```
edwardhorsford:
```suggestion
          <li><a class="govuk-link" href="/drafts?filterSource=Apply">See draft trainees imported from Apply ({{applyDraftCount}})</a></li>
```

app/views/home.html
```diff
@@ -1,156 +1,84 @@
...
+        {% endif %}
+        <li class="govuk-!-margin-bottom-0"><a href="/new-record/new">Add a trainee</a></li>
```
edwardhorsford:
```suggestion
        <li class="govuk-!-margin-bottom-0"><a class="govuk-link" href="/new-record/new">Add a trainee</a></li>
```

app/views/home.html
```diff
@@ -1,156 +1,84 @@
...
-        <div class="govuk-grid-row govuk-!-margin-bottom-6 home-statuses__flex-row">
-          <div class="govuk-grid-column-one-quarter">
-            <a href="/records?filterStatus=EYTS recommended&filterStatus=QTS recommended" class="status-card status-card--recommended">
-              {% set recordCount = records | where("status", ['EYTS recommended','QTS recommended']) | length %}
-              <span class="status-card__count">{{recordCount}}</span>
-              <span class="status-card__status">Qualification recommended</span> <span class="govuk-visually-hidden"> records. View these records.</span>
+    <div class="govuk-grid-column-two-thirds-from-desktop">
+      <h2 class="govuk-heading-l">Registered trainees</h2>
+      <p class="govuk-body">
+        <a href="/records?filterCycle=_unchecked">See all registered trainees ({{ data.filteredRecords | filterByStatus("Draft", true)  | length}})</a>
```
edwardhorsford:
```suggestion
        <a class="govuk-link" href="/records?filterCycle=_unchecked">See all registered trainees ({{ data.filteredRecords | filterByStatus("Draft", true)  | length}})</a>
```

edwardhorsford: I'd create a variable at the top of this section for `nonDraftTrainees` that you can use in the count and later in the various other sections.

app/views/home.html
```diff
@@ -1,156 +1,84 @@
...
+      </p>
+      <h3 class="govuk-heading-m">
+        2020 to 2021 trainees
+        <span class="govuk-body">
+        {# TO DO – add number of records here #}
+           –  <a href="/records?filterCycle=2020+to+2021">see all<span class="govuk-visually-hidden">
+            2020 to 2021 trainees</span> ({{ data.filteredRecords | filterByYear(["2020 to 2021"]) | filterByStatus("Draft", true)  | length}})
```
edwardhorsford: Use filtered nonDraft trainees from above.

app/views/home.html
```diff
@@ -1,156 +1,84 @@
...
+    <div class="govuk-grid-column-two-thirds-from-desktop">
+      <h2 class="govuk-heading-l">Registered trainees</h2>
+      <p class="govuk-body">
+        <a href="/records?filterCycle=_unchecked">See all registered trainees ({{ data.filteredRecords | filterByStatus("Draft", true)  | length}})</a>
+      </p>
+      <h3 class="govuk-heading-m">
+        2020 to 2021 trainees
+        <span class="govuk-body">
+        {# TO DO – add number of records here #}
+           –  <a href="/records?filterCycle=2020+to+2021">see all<span class="govuk-visually-hidden">
```
edwardhorsford:
```suggestion
           –  <a class="govuk-link" href="/records?filterCycle=2020+to+2021">see all<span class="govuk-visually-hidden">
```

app/views/home.html
```diff
@@ -1,156 +1,84 @@
...
-          </div>
-          {% set statuses = ['Deferred', 'Withdrawn'] %}
-          {% for status in statuses %}
-            <div class="govuk-grid-column-one-quarter">
-              {{ statusCard(records, status) }}
-            </div>
+        </span>
+      </h3>
+    </div>
+    <div class="govuk-grid-column-full">
```
edwardhorsford, line 55: This needs to be in a `govuk-grid-row`?

app/views/home.html
```diff
@@ -1,156 +1,84 @@
...
+                </a>
+              {# {% endif %} #}
           {% endfor %}
-        </div>
-      {% endif %}
-
+      </div>
+      <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
+    </div>
+    <div class="govuk-grid-column-two-thirds-from-desktop">
```
edwardhorsford, line 83: This needs to be in a `govuk-grid-row`?

app/views/home.html
```diff
@@ -1,156 +1,84 @@
...
+                {% set statusString = "Qualification recommended" %}
+              {% elif status == 'Awarded' %}
+                {% set searchStatus = ['EYTS awarded', 'QTS awarded'] %}
+                {% set recordCount = records | where("status", ['EYTS awarded', 'QTS awarded']) | length %}
+                {% set statusString = "Qualification awarded" %}
+              {% else %}
+                {% set recordCount = records | where("status", status) | length %}
+            {% endif %}
+            {# {% if recordCount > 0 %} #}
+                <a href="/records?filterStatus={{searchStatus or status}}" class="status-card status-card--{{ status | lower | kebabCase }}">
```
edwardhorsford:
```suggestion
                <a class="govuk-link" href="/records?filterStatus={{searchStatus or status}}" class="status-card status-card--{{ status | lower | kebabCase }}">
```

app/views/home.html
```diff
@@ -1,156 +1,84 @@
...
-      {% endif %}
-
+      </div>
+      <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
+    </div>
+    <div class="govuk-grid-column-two-thirds-from-desktop">
+      <h2 class="govuk-heading-l">Guidance</h2>
+      {# We need to split these pages #}
+      <ul class="govuk-list govuk-list--spaced">
+        <li><a href="/guidance#section-check-data">Find out about Register trainee teachers</a></li>
```
edwardhorsford:
```suggestion
        <li><a class="govuk-link" href="/guidance#section-check-data">Find out about Register trainee teachers</a></li>
```

app/views/home.html
```diff
@@ -1,156 +1,84 @@
...
+        <li><a href="/guidance#section-about-register">Check what data you need</a></li>
```
edwardhorsford:
```suggestion
        <li><a class="govuk-link" href="/guidance#section-about-register">Check what data you need</a></li>
```

edwardhorsford (changes requested):

app/views/home.html
ralph-hawkins: I use it as a way to check if the user has created any manual drafts.

If the total draft count and drafts from apply count are the same then I infer the user has not created any drafts manually and don't show the 'See all draft trainees' link.

app/views/home.html
ralph-hawkins: It is in a `govuk-grid-row`, just one for the whole page. I only use additional `govuk-grid-row`s when I want to control the position of columns at different screen widths. Given that at any width this is always a single column page we don't need the extra control and the whole page can sit in a single `govuk-grid-row`. If we need more columns later we can add more `govuk-grid-row`s.

ralph-hawkins: After feedback from design system team, we're going with more rows is better.

app/views/home.html
edwardhorsford: I meant the line `recordsToBeReviewedCount`

ralph-hawkins: ah — yep

app/views/home.html
```diff
@@ -1,156 +1,87 @@
...
+      </h3>
+    </div>
+  </div>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+      <div class="app-home-statuses">
+
+          {% set statuses = ['Pending TRN', 'TRN received', 'Recommended', 'Awarded', 'Deferred', 'Withdrawn'] %}
+          {% for status in statuses  %}
+          {% set statusString = status %}
```
edwardhorsford: Needs insetting
```suggestion
            {% set statusString = status %}
```

app/views/home.html
```diff
@@ -1,156 +1,87 @@
...
+            {% if status == 'Recommended' %}
+                {% set searchStatus = ['EYTS recommended', 'QTS recommended'] %}
```
edwardhorsford: I think to get the filters working this could do
```suggestion
                {% set searchStatus = "EYTS recommended&filterStatus=QTS recommended" %}
```

app/views/home.html
```diff
@@ -1,156 +1,87 @@
...
+                {% set recordCount = records | where("status", ['EYTS recommended', 'QTS recommended']) | length %}
+                {% set statusString = "Qualification recommended" %}
+              {% elif status == 'Awarded' %}
+                {% set searchStatus = ['EYTS awarded', 'QTS awarded'] %}
```
edwardhorsford: Needs updating to make the filters work correctly.

edwardhorsford (changes requested):

edwardhorsford (approved):

PR #310, opened 2021-07-22
Collect study mode where Publish study mode is ambiguous

edwardhorsford:
Publish allows providers to select `Full time or part time` as the study mode - this means that for some courses we won't know enough to correctly set the study mode.

When selecting a Publish course with an ambiguous study mode, we'll ask a followup question to collect it - in a similar way to subject specialism or itt start date.

This should only apply to manually added trainees with Publish courses with ambiguous study modes. For Apply trainees, providers select the study mode when making the offer - so whilst the publish course might be ambiguous, we'll know for the specific trainee if they're full time or part time.

Screenshot 2021-07-22 at 13 02 48

## Testing

To test, add a trainee on a route that has Publish courses. Select a course that is listed as `full time or part time`. After being asked about specialisms (if subjects are ambiguous) you should be asked about study mode.

The same should work if changing the course from an existing record or Apply draft - though finding a course with ambiguous study mode might be tricky.

### DIFF
--- a/app/views/_includes/forms/course-details/study-mode.html
+++ b/app/views/_includes/forms/course-details/study-mode.html
@@ -1,0 +1,27 @@
+
+{{ govukRadios({
+  fieldset: {
+    legend: {
+      text: pageHeading,
+      isPageHeading: true,
+      classes: "govuk-fieldset__legend--l"
+    }
+  },
+  hint: {
+    text: ""
+  },
+  items: [
+    {
+      text: "Full time"
+    },
+    {
+      text: "Part time"
+    }
+  ]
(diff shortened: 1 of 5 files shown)


app/views/_includes/forms/course-details/study-mode.html
```diff
@@ -0,0 +1,27 @@
...
+  fieldset: {
+    legend: {
+      text: pageHeading,
+      isPageHeading: true,
+      classes: "govuk-fieldset__legend--l"
+    }
+  },
+  hint: {
+    text: ""
+  },
```
ralph-hawkins: I'd remove this as you get an empty div in the page.

ralph-hawkins (approved):
👍🏻

PR #311, opened 2021-07-22
Only ask for study mode for some routes

edwardhorsford:
Study mode does not seem relevant for all routes - this makes it configurable on a route by route basis.

Right now, it's not applied to Assessment only or Opt-in routes, but the exact routes to exclude are still TBC.


ralph-hawkins (approved):
Seems good 👍🏻

PR #313, opened 2021-07-27
What is the trainee’s highest academic qualification? — July 2021

ralph-hawkins:
The previous iteration of the question "What is the trainee’s highest academic qualification?" confused users.

I tried a different approach to see if we could make it work better for users.

localhost_3000_new-record_undergraduate-qualification(screenshot for design history) (2)

In the meantime I also looked at some of the data from DTTP and it doesn't seem to offer much value:

```
72.38%		76 	        Level 3
20.95%		22		Other higher qualification
4.76%		5		Foundation course at FE level
0.95%		1		Advanced Modern Apprenticeships
0.95%		1		Baccalaureate
```

Basically, the majority of users told us exactly what we'd expect, 20% gave us no meaningful data and the rest can't have much statistical significance. So we're going to remove the question entirely, hopefully.

I've added it to the prototype, but hidden it from the trainee creation flows.

### DIFF
--- a/app/assets/sass/application.scss
+++ b/app/assets/sass/application.scss
@@ -201,3 +201,6 @@ span.summary-validation-symbol {
   float: right;
 }
 
+app-nowrap {
+  white-space: nowrap;
+}
--- a/app/views/index.html
+++ b/app/views/index.html
@@ -26,8 +26,8 @@
 
     <h2 class="govuk-heading-m">Pages</h2>
 
-    <p class="govuk-body">Main pages</p>
-    <ul class="govuk-list govuk-list--bullet">
+    <h3 class="govuk-heading-s">Main pages</h3>
+    <ul class="govuk-list">
       <li><a href="start-page" class="govuk-link">Start page</a></li>
       <li><a href="sign-in" class="govuk-link">Sign in</a></li>
       <li><a href="declaration" class="govuk-link">Declaration</a></li>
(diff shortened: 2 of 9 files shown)


app/assets/sass/application.scss
```diff
@@ -201,3 +201,6 @@ span.summary-validation-symbol {
   float: right;
 }
 
+app-nowrap {
```
edwardhorsford, line 204: Nice! 👍 

for later: I've been thinking of making a `utils.scss` file or something - which this would be a good fit for.

app/views/index.html
```diff
@@ -39,22 +39,27 @@ <h2 class="govuk-heading-m">Pages</h2>
...
-    <p class="govuk-body">Error pages</p>
-    <ul class="govuk-list govuk-list--bullet">
+    <h3 class="govuk-heading-s">Error pages</h3>
+    <ul class="govuk-list">
       <li><a href="403" class="govuk-link">400 - Access denied</a></li>
       <li><a href="404" class="govuk-link">404 - Page not found</a></li>
       <li><a href="500" class="govuk-link">500 - Something went wrong</a></li>
     </ul>
 
+    <h3 class="govuk-heading-s">Alternate or unused pages</h3>
```
edwardhorsford, line 58: Nice!

I've not been good at keeping this list up to date - but we should.

I've also been thinking of making a style-guide ish section to link to bespoke patterns for devs - as some of the designs otherwise require complex data set ups.

edwardhorsford (approved):

app/views/index.html
ralph-hawkins: Yeah, that would be super useful.

PR #317, opened 2021-07-29
Swapping between UK and Non-UK degrees

edwardhorsford:
This exposes the answer to UK / Non-UK as an explicit row in the summary list so degrees can be changed type after they're created. Previously, to change the type the user would have had to have deleted the degree and re-added it.

This wasn't so much an issue for manual trainees, but we can see in the Apply data that some applicants may be incorrectly picking UK when they have a non-UK degree. It might not be viable to delete the degree in this case if the Provider doesn't have access to those details.

As well as having a row in the summary card for this, I've also added an explicit toggle from the UK degree details page, which is shown if the institution is invalid - as this is the main time I think someone might want to swap to Non-UK.

## UK degree row on summary cards
Screenshot 2021-07-29 at 15 34 27

## Inset text shown on UK degrees if the institution is invalid.
Screenshot 2021-08-09 at 11 50 24

### DIFF
--- a/app/lib/utils.js
+++ b/app/lib/utils.js
@@ -1468,7 +1468,7 @@ exports.markInput = function(data, params){
   else if (type == 'invalid'){
     message = `${key} is not recognised`
     data.errorMessage = {
-      html: `The trainee entered ‘${valueCleaned}’, which was not recognised. ${params.invalidMessage || 'You need to search for the closest match.'}`
+      html: `The trainee entered ‘${valueCleaned}’${params.invalidMessage || ', which was not recognised. You need to search for the closest match.'}`
     }
     data.classes = (data.classes) ? `${data.classes} app-invalid-answer` : 'app-invalid-answer'
     delete data.value
--- a/app/lib/utils.js
+++ b/app/lib/utils.js
@@ -1443,7 +1443,8 @@ exports.markSummaryRowMissing = function(row) {
 }
 
 
-exports.markInput = function(data, type){
+exports.markInput = function(data, params){
+  let type = params.type
   // Keys are stored two possible places
   let key = data?.label?.html || data?.label?.text || data?.fieldset?.legend?.html || data?.fieldset?.legend?.text
 
@@ -1467,7 +1468,7 @@ exports.markInput = function(data, type){
   else if (type == 'invalid'){
     message = `${key} is not recognised`
     data.errorMessage = {
-      text: `The trainee entered ‘${valueCleaned}’. You need to search for the closest match.`
+      html: `The trainee entered ‘${valueCleaned}’, which was not recognised. ${params.invalidMessage || 'You need to search for the closest match.'}`
     }
     data.classes = (data.classes) ? `${data.classes} app-invalid-answer` : 'app-invalid-answer'
     delete data.value
@@ -1478,7 +1479,7 @@ exports.markInput = function(data, type){
     message = `${key} is missing`
     
     data.errorMessage = {
-      text: message
+      html: message
     }
     data.classes = (data.classes) ? `${data.classes} app-invalid-answer` : 'app-invalid-answer'
     delete data.value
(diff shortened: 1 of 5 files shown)


dankmitchell:
[Add ability to swap between uk and non-uk degrees (S)](https://trello.com/c/TRzmqE4E/2343-add-ability-to-swap-between-uk-and-non-uk-degrees-s)

app/views/_includes/forms/degree/details-international.html
```diff
@@ -10,6 +10,21 @@ <h1 class="govuk-heading-l">{{pageHeading}}</h1>
...
+{% set insetTextHtml %}
+  <p class="govuk-body">Change this to an <a href="./details?degreeTemp[isInternational]=false{{'' | addReferrer(referrer) }}" class="govuk-link">UK degree</a>.</p>
+{% endset %}
+
+{{ govukInsetText({
+  html: insetTextHtml
+}) }} 
+
+#}
+
```
ralph-hawkins, line 26:
```suggestion
{% set degreeCountry = degreeTemp.country %}
{% if degreeCountry == "United Kingdom" %}
  {% set degreeCountry = "" %}
{% endif %}

{% set internationalCountries = data.countries | removeArrayItem("United Kingdom") %}
```

ralph-hawkins: I'm not sure the page should persist "United Kingdom" in the input field if the user has selected that its an international degree.

I don't think "United Kingdom" should appear on the list of countries here either.

ralph-hawkins:
localhost_3000_new-record_degree_0_details_referrer=_new-record_overview,_new-record_apply-trainee-application(screenshot for design history — short)

Happy to go with this, but I'm fairly confident that the option to switch to the a non-UK degree is going to be hard to spot.

It feels like the reading order is wrong too. Ideally the way to change to a non-UK degree should come after the message about why the answer isn't recognised.

app/lib/utils.js
```diff
@@ -1467,7 +1468,7 @@ exports.markInput = function(data, type){
   else if (type == 'invalid'){
     message = `${key} is not recognised`
     data.errorMessage = {
-      text: `The trainee entered ‘${valueCleaned}’. You need to search for the closest match.`
+      html: `The trainee entered ‘${valueCleaned}’, which was not recognised. ${params.invalidMessage || 'You need to search for the closest match.'}`
```
ralph-hawkins: Could the thing its not recognised as be mentioned?

eg – The trainee entered ‘Universite de Lyon 2’, which was not recognised as a university in the UK.

ralph-hawkins (approved):
Very sensible, but I'd be pretty keen to see this tested.

edwardhorsford:
> It feels like the reading order is wrong too. Ideally the way to change to a non-UK degree should come after the message about why the answer isn't recognised.

I originally had a link within the message - but unfortunately this wouldn't work for AT. So putting it before seemed the next best thing.

I *hope* this issue is a rare occurrence - but at least we're providing a forward path for where it is.

app/lib/utils.js
edwardhorsford: @gregknight1 thoughts?

I think if we went with it we'd need to think a bit - 'recognised' possibly implies 'authorised'

app/views/_includes/forms/degree/details-international.html
edwardhorsford: Have done this directly on the field.

ralph-hawkins: nice

app/lib/utils.js
gregknight1: '.....which was not recognised as a UK university' @edwardhorsford @ralph-hawkins

edwardhorsford: @gregknight1 `UK institution`?

gregknight1: @edwardhorsford Yes, that's more inclusive.

PR #320, opened 2021-08-03
Education phase question

edwardhorsford:
Collaboration with @ralph-hawkins 

When adding course details, we'll now ask up front whether it's a primary or secondary course. This means we can offer a better choice of subjects and age ranges as we know what's appropriate. 

We've also seen some participants struggle to complete the form when adding trainees on primary courses. We think this is because they don't instinctively think of those courses as having subjects. By asking about the phase first, we can show a clear question that gets the data we need.

Phase question
Screenshot 2021-08-16 at 14 08 55

Primary course details
Screenshot 2021-08-03 at 17 44 12

If selecting other subject:
Screenshot 2021-08-03 at 17 44 20

The radio question is a shorthand for several inputs - under the hood it still populates 1, 2 or 3 specialisms.

Secondary course details:
Screenshot 2021-08-03 at 17 44 59

Other implementation details:

* Age ranges shown in `other` vary depending on primary or secondary. Where an age range crosses phases, it's shown in secondary only.
* Doesn't apply to Early years as they're essentially all one phase.
* If changing from primary to secondary or visa versa we delete existing subjects and age ranges.
* The two primary type specialisms `Primary teaching` and `Specialist teaching (primary with mathematics)` are no longer shown in the specialisms list for secondary courses.
* When displaying the subject names, I've attempted to preserve what the user picked. So it would say `Primary with geography and history` - even though the specialisms might have slightly different wordings.
* We've changed the age range other away from an autocomplete - is now a regular `select`
* At the moment the Prototype allows changing a publish course's phase - but I wonder if we want to allow that in real life?


app/views/new-record/course-details/level.html
```diff
@@ -0,0 +1,10 @@
+{% extends "_templates/_new-record.html" %}
+
+{% set pageHeading = "What type of course?" %}
```
ralph-hawkins:
```suggestion
{% set pageHeading = "Which stage of education?" %}  
```

I think 'type of course' is too vague. Wikipedia use 'Educational stage' — https://en.wikipedia.org/wiki/Educational_stage — Not sure I love it. @gregknight1 any thoughts?

app/views/_includes/summary-cards/course-details/non-publish-course.html
```diff
@@ -39,6 +39,23 @@
       ]
     } if canAmend
   } if record | routeHasPublishCourses,
+  {
+    key: {
+      text: "Course type"
```
ralph-hawkins:
```suggestion
      text: "Education phase"
```

app/views/_includes/summary-cards/course-details/publish-course.html
```diff
@@ -43,37 +41,37 @@
   } if canAmend and false
 } if record.status !="Draft" %}
 
-{% set subjectsRow = {
+{% set levelRow = {
   key: {
-    text: "Subject" | pluralise(record.courseDetails.subjects | length)
+    text: "Course type"
```
ralph-hawkins:
```suggestion
    text: "Education phase"
```

app/views/new-record/course-details/level.html
edwardhorsford: I'll defer to @gregknight1 on this.

gregknight1: I've mulled this over quite a bit @ralph-hawkins and I get why you think it's vague. We're asking about the course the person will be doing and I felt it important that we referenced that in the heading. But then it gets a bit messy, for example - Which age group is the course to teach? or Which stage of education is the course to teach? Hence why I suggested keeping it vague but allowing the radio labels to do the heavy lifting.

Happy to be persuaded that we don't need to reference the course in the heading and expanding on your suggestion, we could have:

Which education level? 
Which phase of education?

Would also need to be consistent with the heading in the filter too.

ralph-hawkins: You're right it's clear enough from the options, but we have quite a few vague questions. 

For example 'what course are they doing?' and 'what type of training are they doing?').

The main area of benefit i think it brings is the summary page where you can save the user having to scan the value to infer the key. Having said that, this is only a benefit if 'Education stage' is something users recognise.

Not sure about 'level' because of the similarity to 'qualification level'.

Maybe phase is good?

gregknight1: Yeah - really good point about the number of vague questions we've got. Been doing things in Publish today and that has them too. Agree with the 'level' thing - wanted to move away from that

I'm happy for it to be 'phase' ie 'Which phase of education?' then 'Education phase' in the filter and summary.

I think that's a more recognised term than 'stage'.

app/routes/records-list-routes.js
```diff
@@ -165,7 +165,7 @@ const getSelectedFilters = req => {
 
   if (filters.level) {
     selectedFilters.categories.push({
-      heading: { text: 'Course level' },
+      heading: { text: 'Course type' },
```
ralph-hawkins:
```suggestion
      heading: { text: 'Education phase' },
```

app/views/_includes/filter-panel/level.njk
```diff
@@ -2,7 +2,7 @@
   classes: "govuk-checkboxes--small js-auto-submit",
   fieldset: {
     legend: {
-      text: "Course level",
+      text: "Course type",
```
ralph-hawkins:
```suggestion
      text: "Education phase",
```

app/views/new-record/course-details/level.html
ralph-hawkins:
```suggestion
{% set pageHeading = "Which phase of education will the trainee teach?" %}  
```

ralph-hawkins (changes requested):
I think we're agreed on 'Education phase'?

app/views/_includes/summary-cards/course-details/publish-course.html
```diff
@@ -43,37 +41,37 @@
...
+    text: courseDetails.level | sentenceCase or 'Not provided'
   },
   actions: {
     items: [
       {
-        href: recordPath + "/course-details/details" | addReferrer(referrer),
+        href: recordPath + "/course-details/level" | addReferrer(referrer),
         text: "Change",
-        visuallyHiddenText: "subject"
+        visuallyHiddenText: "level"
```
ralph-hawkins:
```suggestion
        visuallyHiddenText: "education phase"
```

app/routes/records-list-routes.js
gregknight1: > I think we're agreed on 'Education phase'?

Yes

ralph-hawkins (changes requested):

ralph-hawkins (approved):
Nice!

PR #325, opened 2021-08-12
Add HPITT training route and example trainees

edwardhorsford:
This adds a new training route to the prototype - HPITT (high potential initial teacher training).

This route is only used by a single provider, so is disabled by default. To see, go to prototype settings and turn the route on.

This adds two seed records for HPITT - one complete draft, and one registered trainee.

The route is basically the same as Assessment Only, but with the addition of a new field - region.

As region is used by a single provider / single user only, initially I'm suggesting just displaying the value and not building an edit flow. I've inserted it in the training details section for now or trainee progress on a registered trainee.

I also took the opportunity to refactor the trainee progress (formerly student record) summary card - which was super complex before.

## Training details
Screenshot 2021-08-12 at 14 48 54

## Trainee progress
Screenshot 2021-08-12 at 14 48 05


dankmitchell:
[Add HPITT to the prototype](https://trello.com/c/hfOHNKZ0/2475-add-hpitt-to-the-prototype)

ralph-hawkins (approved):

PR #326, opened 2021-08-13
Removes 'record' and refers to the trainee

ralph-hawkins:
We can loose most of the uses of 'record'.

The only places I've kept it is to refer to the record as a separate entity from the trainee themselves. For example if the record is missing information (the trainee isn't missing the information).

### DIFF
--- a/app/views/new-record/check-record-apply-grouped-sections.html
+++ b/app/views/new-record/check-record-apply-grouped-sections.html
@@ -1,6 +1,6 @@
 {% extends "_templates/_new-record.html" %}
 
-{% set pageHeading = "Check draft record" %}
+{% set pageHeading = "Check trainee record" %}
 {% set formAction = "./save" | addReferrer(referrer) %}
 {% set hideReturnLink = true %}
 {% set showIncomplete = true %}
@@ -21,7 +21,7 @@
   {% if not recordIsComplete %}
     {% set insetTextHtml %}
       <p class="govuk-body">
-        This trainee record is not complete and cannot be submitted for TRN. If you do not have all the required information now, you can <a href="./save-as-draft" class="govuk-link">return to this draft record later</a>.
+        This trainee record is not complete and cannot be submitted for TRN. If you do not have all the required information now, you can <a href="./save-as-draft" class="govuk-link">return to this draft later</a>.
       </p>
     {% endset %}
       {{ govukInsetText({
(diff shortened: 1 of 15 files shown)


gregknight1:
That all makes sense to me @ralph-hawkins

app/views/new-record/check-record-apply-grouped-sections.html
```diff
@@ -20,7 +20,9 @@
   {# Show a warning if the record isn’t ready to be submitted #}
   {% if not recordIsComplete %}
     {% set insetTextHtml %}
-      <p class="govuk-body">This record is not complete and cannot be submitted for TRN. If you do not have all the required information now, you can <a href="./save-as-draft" class="govuk-link">return to this draft record later</a>.</p>
+      <p class="govuk-body">
+        This trainee record is not complete and cannot be submitted for TRN. If you do not have all the required information now, you can <a href="./save-as-draft" class="govuk-link">return to this draft record later</a>.
```
edwardhorsford: Refers to both `trainee record` and `draft record` on the same line. I think we should be consistent.

app/views/new-record/check-record.html
```diff
@@ -20,7 +20,9 @@
   {# Show a warning if the record isn’t ready to be submitted #}
   {% if not recordIsComplete %}
     {% set insetTextHtml %}
-      <p class="govuk-body">This record is not complete and cannot be submitted for TRN. If you do not have all the required information now, you can <a href="./save-as-draft" class="govuk-link">return to this draft record later</a>.</p>
+      <p class="govuk-body">
+        This trainee record is not complete and cannot be submitted for TRN. If you do not have all the required information now, you can <a href="./save-as-draft" class="govuk-link">return to this draft record later</a>.
```
edwardhorsford: Same as above.

app/views/new-record/check-record-apply-grouped-sections.html
gregknight1: I'd omit the word 'draft' in the 2nd sentence.

ralph-hawkins: I thought remove 'record' so its consistent with the link at the bottom of the page:

localhost_3000_new-record_check-record(screenshot for design history — short) (1)

But tbh I'm not confident. We have 'trainee', 'draft' and 'record'. My preference is to go with draft in most places we have it to emphasise thats its not a 'registered trainee'.

gregknight1: Agree, @ralph-hawkins - I wasn't looking this in context, apologies.

edwardhorsford (approved):

PR #334, opened 2021-09-01
Don't refer to education if no degree section

edwardhorsford:
We don't need to refer to `Personal details and education` when there is no degree section - as we don't ask or display anything about education.


ralph-hawkins (approved):
Nice!

PR #335, opened 2021-09-03
Add concept of salary grants and make bursaries more generic

edwardhorsford:
This PR adds support for collecting information about salary grants - which apply to the `School direct (salaried)` and `Early years (salaried)` routes.

Salary grants are basically the same as bursaries - they have a value `£` and they apply for certain subjects. Mostly what's different is the language.

To do this, I've made the existing bursary functionally more generic in name - it now calls it `financialSupport` and has a type.

## Testing:
* Add a trainee on a `School direct (salaried)` or `Early years (salaried)` route
* Add course details where the allocation subject ends up with one of those that has salary grants (all of early years or `chemistry, computing, maths, physics, languages, classics or biology`)
* When in funding section, the user is now asked about salary grants and is shown the correct value.

Salary grant page (school direct)
Screenshot 2021-09-03 at 14 35 30

Summary card
Screenshot 2021-09-03 at 14 35 39

Salary grant page (early years salaried)
Screenshot 2021-09-03 at 14 37 45


ralph-hawkins:
Looks good — Although I think the question "Are you applying for a salary grant for this trainee?" feels a bit 'informational', like we're asking out of interest. Something like "Do you want to apply for a salary grant for this trainee?" would be clearer that this is the start of the user applying for a grant.

I also think we should add (as a separate story) guidance on what the user can expect from the salary application process on the confirmation page.

cc @gregknight1

edwardhorsford:
> Looks good — Although I think the question "Are you applying for a salary grant for this trainee?" feels a bit 'informational', like we're asking out of interest. Something like "Do you want to apply for a salary grant for this trainee?" would be clearer that this is the start of the user applying for a grant.
> 
> I also think we should add (as a separate story) guidance on what the user can expect from the salary application process on the confirmation page.
> 
> cc @gregknight1

I'll defer to @gregknight1 on thoughts about framing this. I think originally we had some worry that this might not be the method through which applications get made - so having this passive language covered both bases. That still might be the case.

gregknight1:
Will have time to take a look at the content on this tomorrow after chatting to Ed about it this afternoon. @edwardhorsford @ralph-hawkins

ralph-hawkins (approved):

PR #345, opened 2021-09-17
Refactor course details and collect course dates if we don't have them

edwardhorsford:
## Problem

The Publish service doesn't collect full start and end dates for courses - only a start month and an inferred duration. For the moment, Register (and its downstream systems) need full start and end dates.

The first version of Register set the start date to the first of the month of the start month. And the end date to one year from then. So a course starting on September 2021 would get a start date of 1 September 2021 and an end date of 31 August 2022.

We intended that providers would edit this as needed. In practice though, several providers have written in to support to query if they should edit it - they think we might have chosen 31 August intentionally. It's also not ideal that they have to edit the end date for *every* trainee on a course - where the course dates are highly likely to be the ame.

## What this PR does.

* It updates the publish courses to be closer to real publish - they have start months but not full dates.
* When a trainee is being added on to a course without dates, show a conditional page to collect the trainee's course dates
* This page optionally lets the provider save these dates back to the cousre
* If the provider chooses to save those dates, persist those dates back to the course
** NB: courses can be full time *and* part time - and I"m assuming they'll have different dates - so when we save them, we're saving a set of dates per study mode
* When adding more trainees to the same course in the future, we don't need to ask for dates as we'll have them on the course
* If adding a trainee to the same course but with a different study mode, we'll ask for the dates as we won't have them yet.

What we don't support yet:

* There's not any way to update the course's dates once they've been saved. In the future we could have a 'edit course' flow or perhaps a `Courses` section of the service.

## Screenshots
Screenshot 2021-09-28 at 16 47 27

NB: I built an alternate version of the page that _forces_ the saving. It's currently disabled, but for reference:
Screenshot 2021-09-27 at 16 01 24

## How to test

Add a trainee and pick a publish course - the first time you pick that course, it should ask about course dates. If you choose to save the dates, the next trainee you add on that course should get the correct dates.


gregknight1:
@edwardhorsford On the checkbox - with the heading and then the label, it feels a bit repetitive as they're essentially saying the same thing. But do we need to have the heading for accessibility reasons? 

In the checkbox label, I think we're missing the word 'for' between 'dates' and 'all.

Looks nice, though.

ralph-hawkins (approved):

PR #347, opened 2021-09-23
Content update start page

gregknight1:
Experimenting with changelog/what's new type content on the start page - borrowed from the design system after Ralph's suggestion.

- Re-written the summary of the service and moved it into the banner
- Removed the service summary below the banner and included 'Latest updates' content.

**Questions**

How do we handle quite a few updates at once?
How do we handle (or do we need to) historical updates?

Spot the typo!!!!!

### DIFF
--- a/app/views/start-page.html
+++ b/app/views/start-page.html
@@ -61,7 +61,7 @@
 
     <h2 class="govuk-heading-m">Latest updates</h2>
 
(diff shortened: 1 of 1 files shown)


edwardhorsford (approved):

edwardhorsford:
I wonder if some of this could / should also exist when signed in?

PR #350, opened 2021-09-29
Add hints to Provider led

edwardhorsford:
Some providers are unclear about the difference between Provider led and School direct.

Screenshot 2021-09-29 at 10 55 57


gregknight1 (approved):
Good to merge - thanks.

PR #352, opened 2021-09-30
Service updates and patches

ralph-hawkins:
Adds updates to the service.

Home page shows one latest update:

localhost_3000_start-page(screenshot)

Updates page shows all:

localhost_3000_service-updates(screenshot)

Updates to signed-in homepage:

localhost_3000_home(screenshot)

Note — content for updates is placeholder.


app/assets/sass/application.scss
```diff
@@ -221,3 +221,7 @@ span.summary-validation-symbol {
 .app-hidden-form {
   display: inline;
 }
+
+ul {
```
edwardhorsford: I think this should be scoped to a `app-markdown` class or something.

app/data/service-patches-and-updates.json
```diff
@@ -0,0 +1,19 @@
...
+      "title": "Bursaries",
+      "content": "You can record which trainees are applying for a bursary."
+    },
+    {
+      "date": "2021-08-10",
+      "title": "Teaching apprenticeship (postgrad)",
+      "content": "You can now register trainee teachers to Early years (postgrad) training courses. **If** you need help contact"
+    }
+  ]
+}
```
edwardhorsford: Not sure this file is wanted any more.

app/data/service-patches-and-updates.yaml
```diff
@@ -0,0 +1,35 @@
+updates:
```
edwardhorsford: Could the filename be just `service-updates.yaml`? Patches feels rather technical

app/data/session-data-defaults.js
```diff
@@ -162,6 +168,7 @@ module.exports = {
   publishRoutes,
   nonPublishRoutes,
   ukComparableDegrees,
+  updatesAndPatches,
```
edwardhorsford: Would probably call this `serviceUpdates` or `serviceUpdatesAndNews` or something.

app/filters/strings.js
```diff
@@ -71,6 +72,27 @@ filters.currency = input => {
...
+    // },
+    // langPrefix: 'hljs language-', // highlight.js css expects a top-level 'hljs' class.
+    // pedantic: false,
+    // gfm: true,
+    // breaks: false,
+    // sanitize: false,
+    // smartLists: true,
+    // smartypants: false,
+    // xhtml: false
+  })
```
edwardhorsford: Could probably delete all these comments?

app/views/service-updates.html
```diff
@@ -0,0 +1,49 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds-from-desktop">
+        <h1 class="govuk-heading-xl">{{pageHeading}}</h1>
+        <h2 class="govuk-heading-m">Contents</h2>
+        <ul class="govuk-list govuk-list--dash govuk-!-margin-bottom-8 govuk-!-margin-left-3">
+          {% for update in data.updatesAndPatches.updates %}
+            <li><a class="govuk-link" href="#{{ update.title | escape }}">{{ update.title }}</a></li>
```
edwardhorsford: What's this `escape` filter? Don't we want the title camel-cased for the url?

app/data/service-patches-and-updates.yaml
edwardhorsford:
```suggestion
items:
```

app/views/service-updates.html
```diff
@@ -0,0 +1,49 @@
...
+          {% endfor %}
+        </ul>
+
+
+
+        {% for update in data.updatesAndPatches.updates %}
+          <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
+          <section id="section-about-register">
```
edwardhorsford: This id should presumably be the title, camel-cased (there's already a filter to do this).

app/views/service-updates.html
```diff
@@ -0,0 +1,49 @@
...
+            <h2 class="govuk-heading-m govuk-!-margin-bottom-2" id="{{ update.title | escape }}">
```
edwardhorsford: I can see there's an ID on this h2 - wonder if this should be on the parent section instead?

edwardhorsford (changes requested):

gregknight1 (commented):
Looks really good @ralph-hawkins. I take it there's no issue in having a link on the sign-in page. What's springing to mind is how the full list of news/updates is accessed when a user is signed in?

Wondering how we'd approach an update with a large amount of content or multiple updates at the same time.

Also thinking that the contents could end up being massive, as could the actual updates. Think we need some rules around the maintenance of this.

edwardhorsford (approved):

app/views/service-updates.html
```diff
@@ -0,0 +1,38 @@
+
+{% extends "_templates/_page.html" %}
+{% set backLink = '/start-page' %}
```
edwardhorsford, line 3: This should probably be js back - `/start-page` isn't the only source for this page.

PR #354, opened 2021-10-01
Refer to ITT dates rather than course dates

edwardhorsford:
Adopt language around `ITT start date` and `ITT end date` rather than `Course start date` and `Course end date`.

Making this change as although ITT dates and course dates are often the same, for some routes they differ - eg undergrad and teaching apprenticeships. For those routes, asking for `Course start date` was ambiguous - as it could refer to two things - and we're specifically interested in the ITT bit.

### DIFF
--- a/app/data/training-route-data.js
+++ b/app/data/training-route-data.js
@@ -128,8 +128,7 @@ let baseRouteData = {
           ],
         value: "9000"
       }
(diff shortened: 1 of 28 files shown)


ralph-hawkins (approved):
Tidy

PR #360, opened 2021-10-13
Support schools being not applicable

edwardhorsford:
In some cases, a lead and employing school is not applicable - this happens where the trainee is self funded / at a private school and so DfE does not provide funding to the school.

Adds a checkbox to each of the schools pages. When checked, users do not need to provide a school (and we should prevent them providing both).

Lead school page
Screenshot 2021-10-15 at 12 24 33

Employing school page
Screenshot 2021-10-15 at 12 24 47

Confirm page
Screenshot 2021-10-13 at 17 31 13

### DIFF
--- a/app/routes/shared-edit-routes.js
+++ b/app/routes/shared-edit-routes.js
@@ -96,6 +96,12 @@ module.exports = router => {
     let referrer = utils.getReferrer(req.query.referrer)
     const schools = getSchools()
 
+    // default to applicable unless checkbox set
+    let leadSchoolApplicable = true
+    if (record?.schools?.leadSchool?.notApplicable && record?.schools?.leadSchool?.notApplicable.includes('true') ){
+      leadSchoolApplicable = false
+    }
+
     // Input added with js by the autocomplete
     let autocompleteRawValue = req.body?._autocomplete_raw_value_school_picker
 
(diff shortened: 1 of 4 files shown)


ralph-hawkins:
I find 'not applicable' wording a bit hard to parse, I wonder if we can be more specific? @gregknight1 

For example:

```
Trainee is not at a state school or lead school is not listed

If the lead school is missing from the list, contact becomingateacher@digital.education.gov.yk

You only need to provide a lead school if the trainees is employed or funded by a state school

[   ] Lead school is not applicable

```

ralph-hawkins (approved):

gregknight1:
> I find 'not applicable' wording a bit hard to parse, I wonder if we can be more specific? @gregknight1
> 
> For example:
> 
> ```
> Trainee is not at a state school or lead school is not listed
> 
> If the lead school is missing from the list, contact becomingateacher@digital.education.gov.yk
> 
> You only need to provide a lead school if the trainees is employed or funded by a state school
> 
> [   ] Lead school is not applicable
> ```

Yes, I like this @ralph-hawkins, thanks.

@edwardhorsford I'm just checking with Shaun that it's correct to say this - as in, are there any other possible scenarios where this wouldn't be true?

edwardhorsford:
Updated with the suggested wording. Suggest we merge and then we can amend later if needed.

PR #366, opened 2021-10-19
Add badge for incomplete drafts

edwardhorsford:
Adds a badge to the record card to show when the draft is incomplete.

Screenshot 2021-10-19 at 11 00 23

The intention is this would apply to non drafts too - but for the moment the prototype thinks *all* non drafts are incomplete, so this badge would show everywhere.


ralph-hawkins:
I've added rows to the application summary cards. A header row for the name and tags and another body row. It adds a bit more white space, which i think is necessary if we're going to add more to the cards.

Screenshot 2021-10-21 at 14 40 18

ralph-hawkins (approved):

ralph-hawkins (approved):

PR #368, opened 2021-10-20
Show withdrawal reason and date on timeline

edwardhorsford:
We capture withdrawal reason and date, but don't currently display it anywhere in the service.

This adds them as part of the timeline.
Screenshot 2021-10-21 at 11 03 12


app/assets/sass/components/_timeline.scss
```diff
@@ -67,6 +67,11 @@
   margin: 0;
 }
 
+.app-timeline__body-text {
+  @include govuk-font($size: 19);
+  color: $govuk-secondary-text-colour;
```
ralph-hawkins: Does this need to be hint text colour? I'd argue it doesn't fit with the usual reasons we'd make something a hint

ralph-hawkins (approved):

PR #371, opened 2021-10-21
Add reference number

edwardhorsford:
Adds a new field `reference` intended to be a common reference number for trainees. TBC on where we get this - ideally reuse one of the Apply id's, or generate our own for manual applications.

As part of this, I've removed the `trainingDetails` section from the task list - as all it contained was the provider trainee ID. I wouldn't mind letting it be set optionally, but don't feel it warrants a required section if it's optional.

I've updated the seeds so only 30% of records have a trainee id set.

Screenshot 2021-10-21 at 16 49 31


app/views/_includes/summary-cards/trainee-progress.html
```diff
@@ -25,10 +34,20 @@
...
+} %}
+
 
 {% set traineeIdRow = {
   key: {
-    text: "Trainee ID"
+    text: "Provider trainee ID"
   },
   value: {
     text: record.trainingDetails.traineeId or 'Not provided'
```
ralph-hawkins, line 62: 'Not provided' feels a bit strong for an optional answer. It's possible it'll be confused as something the users needs to do to complete a record, especially if we're asking for extra data on records later.

Maybe showing that its optional here too, would help?

app/views/_includes/forms/training-details/trainee-id.html
```diff
@@ -5,7 +5,8 @@
     classes: "govuk-label--l"
   },
   hint: {
-    text: "This will help you to identify them and search for them in your trainee records. We may also use it if we need to get in touch with you about this trainee."
+    text: "This will help you to identify them and search for them in your trainee records. We may also use it if we need to get in touch with you about this trainee.",
```
ralph-hawkins, line 10: I can't see this rendering in the page?

ralph-hawkins (approved):

app/views/_includes/forms/training-details/trainee-id.html
edwardhorsford: I basically made a new line that overwrites this - as I wanted to come back to discuss it before commiting.

PR #382, opened 2021-10-27
Add researchers in schools to PGTA

edwardhorsford:


app/data/training-route-data.js
```diff
@@ -297,6 +297,7 @@ let baseRouteData = {
       // "apprenticeshipStartDate"
     ],
     initiatives: [
+      "Maths and physics chairs programme / Researchers in schools",
```
ralph-hawkins, line 300: Could this be an "and" or "or" instead of a slash?

edwardhorsford: I think they're two names for the *same* thing.

ralph-hawkins: "also know as"?

PR #386, opened 2021-11-01
Refactor autocompletes

edwardhorsford:
This pr:
* Makes a new autocomplete component which uses more sensible code.
* The component calls a `govukInput` or `govukSelect` with appropriate metadata attached
* A separate js file initialises any autocompletes it finds on the page
* I've updated the degree autocompletes to use this - though haven't yet deleted the old autocomplete code

I've also updated to use the new institutions list with expanded synonyms.

## Updated styling for suggestions.
When `showSuggestionsBanner` is true, the menu is prefaced with `Suggestions`. This is removed when there are no more results. This follows the designs used by ONS in the census.

Screenshot 2021-11-11 at 15 04 38

Screenshot 2021-11-11 at 15 04 47

Video
autocomplete-with-suggestions3


ralph-hawkins:
Because the suggestion field is as wide as the input it looks a lot like a dropdown.

Screenshot 2021-11-04 at 13 48 59

ONS makes them different lengths, I think this really helps to make the suggestion feel less like a dropdown.

Screenshot 2021-11-04 at 14 36 07

ralph-hawkins (approved):

PR #394, opened 2021-11-12
adds org as caption to records and drafts

ralph-hawkins:
I think this should be used across all the pages that are the first page from the top nav

Screenshot 2021-11-12 at 11 36 12

Screenshot 2021-11-12 at 11 36 22


edwardhorsford (approved):

edwardhorsford:
Note - what should it show for admins?

PR #401, opened 2021-11-29
Multiple providers and lead schools

edwardhorsford:
This expands on the previous work on multiple providers to add the ability to be associated with a lead school.

As part of this, I've added support for authorisation so that we can show or hide things depending on the auth level of the provider the viewing users is acting as.

We've decided to go with the hat model, but for now the prototype continues to support both models.

For lead school views, mostly things have been hidden:
* No viewing of drafts
  * Not mentioned on homepage
  * Drafts tab removed
* No ability to add trainees
* No ability to export (tbc?)
* Accessing records
  * Can view records
  * Cannot edit records
  * disability and diversity section is hidden
  * gender and nationality is hidden
  * Cannot recommend / delete/ defer / withdraw
  * Accrediting provider row always shown
* Accrediting provider shown in record card 
* Incomplete records
  * Incomplete tags not shown to lead schools
  * Incomplete filters not shown to lead schools 

Not done / things to think about for the future:
* Should we have a filter for accrediting provider so lead schools can scope their work?
* Provider switcher page probably needs work
* Should we allow exporting?
* How do users get access to lead schools?

### Lead school homepage
Screenshot 2021-11-29 at 14 46 20

### Records list
Screenshot 2021-11-29 at 14 50 46

### Record view
Screenshot 2021-11-29 at 14 51 11


ralph-hawkins:
Looks good

A few things:
- consider adding something on the lead school page so its clear this will allow them to see a registered trainee
- looks like a there's a bug with the code that hides TRNs for records that don't have TRNs. They now show as undefined
- should we hide gender?
- the blue incomplete pattern feels weird to show to lead schools, given they can't do anything
- the incomplete tag also feels weird to show to a lead school for the same reason
- perhaps some of the guidance needs rewording to frame what a SCITT does?

edwardhorsford:
[Prototype: Add support for lead schools](https://trello.com/c/Ja0tyMgh/3250-prototype-add-support-for-lead-schools)

edwardhorsford:
[Prototype: Add support for multiple providers](https://trello.com/c/iO8rQBwE/3251-prototype-add-support-for-multiple-providers)

edwardhorsford:
Done:
- [x] looks like a there's a bug with the code that hides TRNs for records that don't have TRNs. They now show as undefined
- [x] should we hide gender?
- [x] the blue incomplete pattern feels weird to show to lead schools, given they can't do anything

Screenshot 2021-11-30 at 17 27 34

To discuss with @gregknight1 
- [x] consider adding something on the lead school page so its clear this will allow them to see a registered trainee
- [x] the incomplete tag also feels weird to show to a lead school for the same reason
- [ ] perhaps some of the guidance needs rewording to frame what a SCITT does?

edwardhorsford:
@ralph-hawkins I've addressed everything except the guidance bit, which @gregknight1 will do as a separate PR.

ralph-hawkins (approved):

PR #405, opened 2021-12-02
Link between draft and registered trainee search results

edwardhorsford:
We saw in user research a user search for a user by using the search input for the user's name. However the record they were looking for was for an existing, registered trainee, and they were in the drafts page. As a result they got no results.

The drafts page and registered trainee pages both look similar. A consequence / downside of having scoped pages like this is you might be on the wrong one.

This tries to help a little by adding a link to the other page if there are results there.

Screenshot 2021-12-02 at 17 54 41

Notes about the language:
* Multiple bits are pluralised depending on if there are 1 result or more.
* If there are results on the current page, then the word `more` is added - `There is 1 result` > `There is 1 more result`

### DIFF
--- a/app/views/records.html
+++ b/app/views/records.html
@@ -167,6 +167,21 @@
     <h2 class="govuk-heading-m">No records found</h2>
     <p class="govuk-body">Try <a href="/records">clearing your search and filters</a>.</p>
   {% endif %}
+
+  {% if draftRecordsCount and isAuthorised('viewDrafts') %}
+    {% set insetHtml %}
+      <p class="govuk-body">
+        <a href="/drafts{{queryString}}">
+          There {{ "are" | pluralise(draftRecordsCount) }} {{draftRecordsCount}}{{ " more" if filteredRecords | length}} {{ "result" | pluralise(draftRecordsCount) }} in your registered trainees.
+        </a>
+      </p>
+    {% endset %}
+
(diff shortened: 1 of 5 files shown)


app/views/records.html
```diff
@@ -167,6 +167,21 @@ <h2 class="govuk-heading-m">Records</h2>
     <h2 class="govuk-heading-m">No records found</h2>
     <p class="govuk-body">Try <a href="/records">clearing your search and filters</a>.</p>
   {% endif %}
+
+  {% if draftRecordsCount and isAuthorised('viewDrafts') %}
+    {% set insetHtml %}
+      <p class="govuk-body">
+        <a href="/drafts{{queryString}}">
+          There {{ "are" | pluralise(draftRecordsCount) }} {{draftRecordsCount}}{{ " more" if filteredRecords | length}} {{ "result" | pluralise(draftRecordsCount) }} in your registered trainees.
```
ralph-hawkins:
```suggestion
          There {{ "are" | pluralise(draftRecordsCount) }} {{draftRecordsCount}}{{ " more" if filteredRecords | length}} {{ "result" | pluralise(draftRecordsCount) }} in your draft trainees.
```

ralph-hawkins (changes requested):
Small typo — but looks good

ralph-hawkins (approved):

PR #412, opened 2021-12-08
Add filters for undergrad / postgrad

edwardhorsford:
Explicitly sets that all routes are either undergraduate or postgraduate.

Adds a filter for undergraduate / postgraduate on the basis of the route set.

Adding this in preparation of doing different flows depending on if the trainee is undergrad or postgrad.

Screenshot 2021-12-08 at 16 59 41


ralph-hawkins (approved):
Looks good

PR #416, opened 2021-12-10
updates the content for the update

ralph-hawkins:


edwardhorsford (commented):
My previous comment on this was that 'start date' was ambiguous - I think this could be phrased more clearly.

app/data/service-updates.yaml
```diff
@@ -3,9 +3,11 @@
 items:
 
   - date: '2021-12-10'
-    title: How Register collects initial teacher training (ITT) start dates has changed
+    title: When you need to provide trainee start dates has changed
     content: >-
-      When registering a trainee, we were always asking for an estimated start date. Now, we will only ask for a start date after a trainee has started.
+      When registering a trainee, we were always asking for a start date. Now, we will only ask for a start date after the ITT has started.
```
gregknight1: When registering a trainee, we were always asking for a start date, even if the course had not started. Now, we will only ask for a start date once the course has started.

app/data/service-updates.yaml
```diff
@@ -3,9 +3,11 @@
 items:
 
   - date: '2021-12-10'
```
gregknight1: 2022-01-20

gregknight1 (changes requested):

app/data/service-updates.yaml
ralph-hawkins: I think we need to be more explicit about which date we're asking for:

> When registering a trainee, we were always asking ~~for a start date~~ `when the trainee started`, even if their course had not started. Now, we will only ask ~~for a start date once the~~ `when the trainee started, after their`  course has started.

gregknight1: Yes, agree with that @ralph-hawkins

gregknight1:
Looks good @ralph-hawkins

PR #423, opened 2021-12-16
Defer withdraw check new start date

ralph-hawkins:
this pr contains some stuff from that got reverted:
- using 'trainee start date' not 'trainee’s start date'
- bug that meant heading for when did the trainee start didn't change with context

adds
- update to change flow for defer/withdraw — if the user sets a new trainee start date that is after the deferral/withdrawal date then they will be asked for the deferral/withdrawal date again

### DIFF
--- a/app/filters/dates.js
+++ b/app/filters/dates.js
@@ -170,6 +170,31 @@ filters.toDateArray = (date) => {
 
 /*
   ====================================================================
+  toDateObject
+  --------------------------------------------------------------------
+  Convert an array to date
+  ====================================================================
+
+  Usage:
+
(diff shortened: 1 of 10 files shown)


app/filters/dates.js
```diff
@@ -168,6 +168,30 @@ filters.toDateArray = (date) => {
...
+
+  Usage:
+
+    {{ "1 January 2021" | toDateArray }}
+
+  = [1, 1, 2021]
+
+*/
+
+// Output date array - for use in design system macros macros
```
edwardhorsford: These docs are wrong

edwardhorsford (approved):

PR #436, opened 2022-01-21
Delete is called remove

ralph-hawkins:
We saw that users were nervous about deleting trainees.

This re-frames 'deleting' as a thing to do when the trainee does not start their course and calls it 'remove'.

Because the remove/delete flow is initiated by the user stating the trainee did not start, we can remove the pages asking if the trainee started where this is ambiguous.

localhost_3000_record_34dc2e7c-4123-4ef4-ad7e-ab9b0dbe54e8(screenshot)

localhost_3000_record_34dc2e7c-4123-4ef4-ad7e-ab9b0dbe54e8(screenshot) (1)

localhost_3000_record_e432974c-e719-4ab0-96a2-a144532cec80(screenshot)

localhost_3000_record_e432974c-e719-4ab0-96a2-a144532cec80_delete_confirm_referrer=_record_e432974c-e719-4ab0-96a2-a144532cec80(screenshot)

Example mobile width:

localhost_3000_record_e432974c-e719-4ab0-96a2-a144532cec80(iPhone SE) (6)

If the user has javascript disabled they see up to 3 buttons:

no-js

No js at mobile width

localhost_3000_record_e432974c-e719-4ab0-96a2-a144532cec80(iPhone SE) (1)


gregknight1:
'Change attendance' feels a little strange – maybe, 'Change status'

Withdraw and Defer are both statuses in Register. And you could argue 'Did not start' is a status too, but now they can be removed so you wouldn't see it.

Feels like we could give more context on the 'Remove this record' red button page. For example:

You can remove this record because the trainee has not started. Removing a trainee means the record will no longer appear in Register. If the trainee returns to training, you will need to create a new record. (_not those exact words but just giving an example of what the implications are for the user if they choose to remove_)

ralph-hawkins:
Thanks @gregknight1 

> 'Change attendance' feels a little strange – maybe, 'Change status'

You're right. I don't love ‘attendance’ but I'm not sure if ‘status’ isn't too vague? Happy to put it in front of users.

> Feels like we could give more context on the 'Remove this record' red button page. For example:

I've added: 

localhost_3000_record_f202cc6d-3a97-4353-be40-6cd5f3b0a6d9_remove_confirm_referrer=_record_f202cc6d-3a97-4353-be40-6cd5f3b0a6d9(screenshot)

gregknight1:
Looks good.

Yes - let's put the attendance/status thing in front of users.

app/assets/sass/application.scss
```diff
@@ -244,3 +244,15 @@ $sizes: 40, 35, 20, 15, 10;
...
+  .moj-button-menu__item { 
+    border-collapse: collapse;
+    border-left:   1px solid $govuk-border-colour;
+    border-right:  1px solid $govuk-border-colour;
+    border-bottom: 1px solid $govuk-border-colour;
+    &:focus {
+      border: none;
+    }
+  }
+}
```
edwardhorsford: This should probably be in a vendor file for the moj button menu?

edwardhorsford (approved):

PR #439, opened 2022-01-28
Improvements to placements

edwardhorsford:
A few bits of clean up to placements:

* Bespoke fail page when you add the same placement twice
* Fix bugs
* Show summary card when there are no placements (this happened sometimes before but was not consistent)

## Bespoke fail page
Adds a bespoke fail page where a user attempts to add a placement that matches an existing one.

Screenshot 2022-01-28 at 11 34 28

### DIFF
--- a/app/views/_includes/summary-cards/placements/placement-overview.html
+++ b/app/views/_includes/summary-cards/placements/placement-overview.html
@@ -29,7 +29,7 @@
           visuallyHiddenText: (loop.index | getOrdinalName | sentenceCase) + "placement"
         }
       ]
-    } if canAmend and false
+    } if canAmend
   }) %}
 
   {% if placements | length == 1 %}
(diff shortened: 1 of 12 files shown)


app/views/_includes/summary-cards/placements/placement-overview.html
```diff
@@ -101,7 +101,7 @@
     actions: {
       items: [{
         href: recordPath + "/placements/confirm" | addReferrer(referrer),
-        text: "Change" if canAmend else "View"
+        text: "Manage placements" if canAmend else "View"
```
ralph-hawkins, line 104: I don't think we need the 'View', not sure it adds anything for the user

ralph-hawkins (approved):
Nice!

app/views/_includes/summary-cards/placements/placement-overview.html
edwardhorsford: `Manage` felt weird where you don't have edit rights or the item can't be edited.

ralph-hawkins: I'd advocate we hide it. The user can't see anything on this page they can't see on the record overview.

edwardhorsford: Fair enough. It was a bit of a legacy from previously where you _could_ see more information on this page.

I might just hide for now as I think we may have further placement information to collect.

PR #443, opened 2022-02-03
Bugfixes feb

edwardhorsford:
Fixes:
* Not enough seed records being generated for Coventry University
* Placements crashing on one of the seed records
* Trainees with complete placements would still show as 'incomplete' sometimes
* Pointers to drafts or records for other search results were showing even when filters not applied
* Suppresses back link on organisation selection page (as agreed with devs)
* Changes default org to a SCITT


edwardhorsford:
[Prototype bugfixes](https://trello.com/c/SizoQWn6/3614-prototype-bugfixes)

scripts/generate-records.js
```diff
@@ -201,7 +201,7 @@ const generateFakeApplications = () => {
     yearsToGenerate.forEach((year) => {
       // Years can be ±10% in size
       let traineeCount = getRandomArbitrary((providerSize * 0.9), (providerSize * 1.1))
-      if (provider == "Coventry University") traineeCount = 130
+      if (provider?.name == "Coventry University") traineeCount = 130
```
ralph-hawkins: Should this be `Webury Hill SCITT`?

ralph-hawkins (approved):

scripts/generate-records.js
edwardhorsford: It is! good catch

PR #445, opened 2022-02-08
Funding in monthly tables inside accordions

ralph-hawkins:
We have an unknown data set coming from the funding team, so I wanted to make sure we could handle any complexity of data.

To do this we've decided to show a simple breakdown (payment and running total per month) and then show a detailed breakdown for each month. 

localhost_3000_funding_monthly-payments(screenshot) (9)

### DIFF
--- a/app/assets/sass/application.scss
+++ b/app/assets/sass/application.scss
@@ -226,7 +226,7 @@ span.summary-validation-symbol {
 
 // column widths for tables
 // app-table__column-xx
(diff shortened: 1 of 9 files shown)


edwardhorsford (approved):
Looking good!

PR could do with a bit of a narrative description for the changes.

I think we should consider showing the year alongside the months - and if not that, then at least for January months.

Would also consider introducing August to the accordions so there's parity with them and the table.

PR #453, opened 2022-02-18
HESA records

ralph-hawkins:
Register is importing HESA records. These records will be view only until the last update from HESA is imported.

Changed proposed:
- adds a check to see if a record can be edited
- makes records that should be locked view only (similar to lead schools — hides incomplete badges and change links)
- adds an explanation about the record being view only

localhost_3000_record_92f655b3-e0f1-4a8e-96be-c1fdd59410e8(screenshot)


app/filters/strings.js
```diff
@@ -100,6 +100,22 @@ filters.startsWith = (string, target) => {
...
 }
 
+// Make a string possessive
+// {{ "James Joyce" | possessive }} 
+//     James Joyce’s
+// {{ "Joyce James" | possessive }} 
+//     Joyce James’
+
+filters.possessive = (string) => {
+  lastLetterOfString = string.split('').slice(-1)
```
edwardhorsford: you should lowercase the string.

app/lib/utils.js
```diff
@@ -948,6 +952,48 @@ exports.isPreviousYears = record => {
...
+
+  const endAcademicYear = moment(record?.courseDetails?.endDate).format("YYYY")
+  const finishedPreviousYear = (endAcademicYear < years.endOfCurrentCycle)
+  const isFinishingThisYear = (endAcademicYear == years.endOfCurrentCycle)
+
+  const dateThisRecordUnlocksOn = exports.dateHesaRecordUnlocked(record)
+
+  let shouldBeLocked = false
+
+  if (source == "HESA") {
```
edwardhorsford: i'd use your `sourceIsHesa` filter here so we only have one way of checking this.

app/lib/utils.js
```diff
@@ -948,6 +952,48 @@ exports.isPreviousYears = record => {
...
+  - or its finishing this year, but todays date is before the cut-off
+*/
+
+exports.isHesaAndLocked = record => {
+
+  const source = record?.source
+
+  const endAcademicYear = moment(record?.courseDetails?.endDate).format("YYYY")
+  const finishedPreviousYear = (endAcademicYear < years.endOfCurrentCycle)
+  const isFinishingThisYear = (endAcademicYear == years.endOfCurrentCycle)
```
edwardhorsford: Maybe not for this PR, but we should consider these as generic filters.

app/views/_includes/summary-cards/placements/placement-overview.html
```diff
@@ -101,7 +101,7 @@
     actions: {
       items: [{
         href: recordPath + "/placements/confirm" | addReferrer(referrer),
-        text: "Manage placements" if canAmend else "View"
+        text: "Manage placements" if canAmend
```
edwardhorsford: Is this change intentional? I'm not entirely sure about 'manage' when you don't have edit rights.

app/views/_templates/_record.html
```diff
@@ -75,6 +75,28 @@ <h1 class="govuk-heading-xl govuk-!-margin-bottom-0">
...
+    {% endset %}
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds-from-desktop">
+        {{ govukInsetText({
+          text: hesaRecordSummary | safe,
+          classes: "govuk-!-padding-top-0 govuk-!-padding-bottom-0"
+        }) }}
+      </div>
+    </div>
+  {% endif %}
```
edwardhorsford: Thoughts on this going in to an include?

app/views/_templates/_record.html
```diff
@@ -75,6 +75,28 @@ <h1 class="govuk-heading-xl govuk-!-margin-bottom-0">
...
+  {% if record | isHesaAndLocked %}
+    {% set hesaRecordSummary %}
+      <h3 class="govuk-heading-s">
+        This record was imported from HESA and cannot be edited
+      </h3>
+      <p class="govuk-body">
+        Updates from HESA are imported regularly.
+      </p>
+      <p class="govuk-body">
+        You will be able to recommend the trainee for QTS at the end of the academic year.
```
edwardhorsford: HESA is never EYTS, right?

edwardhorsford (changes requested):

app/filters/strings.js
ralph-hawkins: Good point, I've made it handle upper case and lower case instead of forcing the string to lowercase.

app/views/_includes/summary-cards/placements/placement-overview.html
ralph-hawkins: This hides "View" if the user doesn't have edit rights. I don't think the secondary view adds anything here as the content is the same and its unnecessary complexity in the (prod) codebase.

app/views/_templates/_record.html
ralph-hawkins: sensible

app/views/_templates/_record.html
ralph-hawkins: Yeah, pretty sure

app/views/_includes/summary-cards/placements/placement-overview.html
edwardhorsford: ah - in which case we should put a condition on the actions section entirely - I think this will output an anchor with empty text?

ralph-hawkins: Ah, yeah — good shout

app/lib/utils.js
ralph-hawkins: Done

app/filters/strings.js
```diff
@@ -100,6 +100,30 @@ filters.startsWith = (string, target) => {
...
+//     James Joyce’s
+// {{ "Joyce James" | possessive }}
+//     Joyce James’
+// {{ "JAMES JOYCE" | possessive }}
+//     JAMES JOYCE’S
+// {{ "Joyce James" | possessive }}
+//     JOYCE JAMES’
+
+filters.possessive = (string) => {
+  const lastLetterOfString = string.split('').slice(-1).toString()
```
edwardhorsford:
```suggestion
  const lastLetterOfString = string.split('').slice(-1).toString().toLowerCase()
```

Just convert this to lowercase and then later compare to lowercase `s`

app/filters/strings.js
```diff
@@ -100,6 +100,30 @@ filters.startsWith = (string, target) => {
...
+  if (lastLetterOfString == "s" || lastLetterOfString == "S") {
+    return string + "’"
```
edwardhorsford:
```suggestion
  if (lastLetterOfString == "s") {
    return string + "’"
```

app/filters/strings.js
```diff
@@ -100,6 +100,30 @@ filters.startsWith = (string, target) => {
...
+  } else if (lastLetterOfString == lastLetterOfString.toUpperCase()) {
+    return string + "’S"
```
edwardhorsford:
```suggestion
```

Don't think we need this? would we ever want capitalised possession?

app/filters/strings.js
```diff
@@ -100,6 +100,30 @@ filters.startsWith = (string, target) => {
...
+  } else if (lastLetterOfString == lastLetterOfString.toLowerCase()) {
+    return string + "’s"
```
edwardhorsford: Unsure what this is testing.

app/lib/utils.js
```diff
@@ -948,10 +952,58 @@ exports.isPreviousYears = record => {
...
+     } else if (exports.isFutureYear(record)) {
+        shouldBeLocked = true
+     }
+     else if (exports.isFinishingThisAcademicYear(record) && (moment().isBefore(exports.dateHesaRecordUnlocked(record)))) {
+        shouldBeLocked = true
+     }
+   } 
+  else {
+    shouldBeLocked = false
+  }
```
edwardhorsford: You don't really need this since the default is for it to be false.

edwardhorsford (changes requested):

app/lib/utils.js
ralph-hawkins: Removed

edwardhorsford (approved):

PR #457, opened 2022-02-24
Pick routes using more realistic ratios

edwardhorsford:
The current seed generators picks routes randomly. Whilst this gives us a spread of routes, it means the niche routes are *way* over represented.

This adds some approximate weighting to the routes, split by SCITT and HEI - so we should see more realistic numbers.

### DIFF
--- a/app/data/generators/route.js
+++ b/app/data/generators/route.js
@@ -1,0 +1,71 @@
+const faker = require('faker')
+const moment = require('moment')
+const weighted = require('weighted')
+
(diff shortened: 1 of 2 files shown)


edwardhorsford:
[Prototype: adjust weighting of routes to be more realistic](https://trello.com/c/guc9Adzb/3730-prototype-adjust-weighting-of-routes-to-be-more-realistic)

ralph-hawkins (approved):

PR #462, opened 2022-03-03
Support changing route and course on records

edwardhorsford:
This pr makes a bunch of changes around routes and courses.

The primary driver is to support changes of routes and courses on registered trainees. Right now when changing course, we only show courses for the current route - so we want to enable changing routes. At the same time, not all routes have courses in the first place.

The solution is that when changing a course, we *always* ask the route first, then follow up with an appropriate course page. So if the route has no courses, we show a manual course details flow. If a route does have courses, we show a course picker flow.

This basically means that picking the route becomes question 1 of course details. We treat manual drafts a little special in that we ask this question out of sequence from the rest of course details.

## Key changes:

* New omni course row in course details

This row shows the route as well as the Publish course name if there is one. If there's no course, then it just shows the route. Selecting change takes the user through a route selection page, then course selection or course details
Screenshot 2022-03-03 at 15 46 24

* When confirming the course on an Apply application, if the user opts to change, we ask about route.
* This means the course picker for apply applications no longer needs to show all Publish courses, but can show just those for the selected route - making this flow closer to the manual drafts flow.
* When changing route or course, we clear out select bits of data that are incompatible with the current choice, but retain what might be the same (this might be worth a discussion on the merits of doing this)


ralph-hawkins (approved):

ralph-hawkins (commented):
i think there might be a case for always clearing dates if you go to/from AO or EYAO as the course dates likely to be significantly different from the others

app/views/_includes/summary-cards/course-details/non-early-years-course.html
```diff
@@ -8,30 +8,60 @@
...
+    <p class="govuk-body">{{ courseDetails.courseNameLong }}</p>
+    {# <p class="govuk-body">{{ courseDetails.route }}</p> #}
+  {# {% else %}
+    <p class="govuk-body">{{ courseDetails.subjects | prettifySubjects or 'Not provided' }}</p> #}
+     {# (course details added manually) #}
+  {% endif %}
+
+{% endset %}
+
+{% set omniCourseRow = {
```
ralph-hawkins: Perhaps this could be called 'CourseAndRoute'?, 'omniCourse' might be a bit hard to decipher if anyone new joins.

PR #463, opened 2022-03-10
Collect extra needed information when changing course or route

edwardhorsford:
When a user makes a significant change to course details (subject 1 or changing route) we should:
* Confirm the key training details including funding
* Collect any extra information now needed (because they've changed to a route that needs more info)

This PR adds a new flow that's triggered when a significant course change is made. After the user confirms the course change, we show an interstitial about the flow, then if there's new information needed, we proceed in turn to collect that information. Once the information is provided, we replay the trainee data. 

Key new pages:
* Interstitial page
* Review page
* Cancel page

## Interstitial page
Most cases when going down this flow, data will be needed - in this case the page tells the user what's going to happen
Screenshot 2022-03-16 at 10 47 03

In some cases there is no new data needed, so we just talk about reviewing the data:
Screenshot 2022-03-16 at 10 50 33

## Review page

This shows a minimal summary of the course, and if applicable, degree, schools, funding

Screenshot 2022-03-16 at 15 17 30

## Cancel page

If a user tries to cancel past the point of confirming the course, we show an interstitial cancel page informing them that cancelling will cancel the course changes too.

Screenshot 2022-03-16 at 10 59 56


edwardhorsford:
[Backfill schools and funding when changing courses](https://trello.com/c/yrniU7kb/3802-backfill-schools-and-funding-when-changing-courses)

app/routes/shared-edit-routes.js
```diff
@@ -834,6 +848,65 @@ module.exports = router => {
...
+
+    let previousRecord = utils.getRecordById(data.records, record.id)
+
+    let routeChanged = (previousRecord?.route != record?.route)
+
+    let firstSubject = record?.courseDetails?.subjects?.first
+    let previousFirstSubject = previousRecord?.courseDetails?.subjects?.first
+    let firstSubjectChanged = !Boolean(firstSubject) || (previousFirstSubject !== firstSubject)
+
+    // Clear out school data if schools no longer needed
```
ralph-hawkins, line 867: Not sure this hint is quite right?

ralph-hawkins (approved):

PR #467, opened 2022-03-17
Optimise prototype

edwardhorsford:
Makes several changes to limit crashes due to lack of memory and reduce unnecessary processing.


edwardhorsford:
[Prototype: investigate performance issues](https://trello.com/c/ugJs5oHY/3831-prototype-investigate-performance-issues)

ralph-hawkins (approved):
Seems a lot better to me. Is it worth asking a dev to review?

PR #472, opened 2022-03-23
Move error pages and update them

edwardhorsford:
Move the error pages in to a folder so we can add more without cluttering the file system.

Also updates the content and email addresses and removes the nav on the one that wouldn't have nav.


ralph-hawkins (approved):
we could make the email a filter that takes an input for the mailto?

PR #477, opened 2022-03-31
Update privacy-policy.html

MaeveRoseveare:
- Updated privacy policy to match production and add tweaks from data protection team

### DIFF
--- a/app/views/privacy-policy.html
+++ b/app/views/privacy-policy.html
@@ -2,155 +2,157 @@
 {% extends "_templates/_page.html" %}
 {% set backLink = '/home' %}
 
-{% set pageHeading = 'How we use your personal data and trainee data in Register trainee teachers' %}
+{% set pageHeading = 'Register trainee teachers privacy policy' %}
 
 {% block content %}
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds-from-desktop">
-    <span class="govuk-caption-l">Privacy policy</span>
     <h1 class="govuk-heading-l govuk-!-margin-bottom-9">{{pageHeading}}</h1>
 
-    <p class="govuk-body">This service allows you to share and access trainee data with the Department for Education (DfE).</p>
+    <h2 class='govuk-heading-m' id='who-are-we'>Who we are</h2>
+​
+      <p class='govuk-body'>Register trainee teachers (‘Register’) is run by the Department for Education (‘we’) for the purpose of registering trainee teachers.</p>
+​
+      <p class='govuk-body'>For the purpose of UK General Data Protection Regulation (UK GDPR), the DfE is the data controller for data held and processed in Register. Initial teacher training (ITT) providers, who use Register, are both data processors for DfE and also independent Data Controllers within the delivery of teacher training.</p>
+​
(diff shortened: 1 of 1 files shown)


MaeveRoseveare:
[Add new privacy policy to prototype](https://trello.com/c/9B7S9Jz4/3874-add-new-privacy-policy-to-prototype)

app/views/privacy-policy.html
```diff
@@ -2,155 +2,157 @@
...
+        <li>name</li>
+        <li>email address</li>
+      </ul>
+​
+      <h3 class='govuk-heading-s'>The data we collect from trainee teachers</h3>
+      <p class='govuk-body'>We collect your personal data to register you with the DfE as a trainee teacher. We collect your:</p>
+      <ul class='govuk-list govuk-list--bullet'>
+        <li>first name</li>
+        <li>middle name</li>
+        <li>last name</li>
```
ralph-hawkins: Do we need to break these down like this? Could it just be "Your full name"?

app/views/privacy-policy.html
```diff
@@ -2,155 +2,157 @@
...
+        <li>gender</li>
```
ralph-hawkins: Do we mean gender? @edwardhorsford has changed the question to 'sex', given what we're asking and how it's being asked.

app/views/privacy-policy.html
```diff
@@ -2,155 +2,157 @@
...
+      <p class='govuk-body'>The personal data we collect will depend on whether you’re a trainee teacher or an ITT provider staff member.</p>
+      <p class='govuk-body'>We will state in this privacy policy when we’re referring to trainee personal data and when we’re referring to ITT provider staff personal data.</p>
+​
+      <h3 class='govuk-heading-s'>The data we collect from ITT provider staff</h3>
+​
+      <p class='govuk-body'>We use your personal data to run and improve Register. We collect your:</p>
+      <ul class='govuk-list govuk-list--bullet'>
+        <li>name</li>
+        <li>email address</li>
+      </ul>
```
ralph-hawkins: Should this section on data we collect about ITT provider staff be scoped to the people who have accounts? We don't collect anything on ITT provider staff otherwise.

ralph-hawkins (approved):
Looks good!

## Mark-up changes

There's a couple of things I'd change in the mark-up, [I've created a PR](https://github.com/DFE-Digital/register-trainee-teachers-prototype/pull/478) I can merge into this, if you're happy?

## Minor feedback on the content.

I wonder if it'd be a little easier to read if the sections on what data is collected and how its used was grouped by persona? So the data we collect about ITT provider staff is grouped with how its used and the same for trainee teachers.

app/views/privacy-policy.html
MaeveRoseveare: Sounds good, I can reshuffle things. Also, yes happy for you to merge your pr.

app/views/privacy-policy.html
MaeveRoseveare: Ah right, that might be a legacy thing and I wasn't aware of the change. Thanks for flagging!

PR #478, opened 2022-03-31
mark-up changes to privacy policy

ralph-hawkins:
- removes white space (this isn't picked-up by github, but there was extra line breaks between paragraphs)
- makes single quotes into double quotes


MaeveRoseveare (approved):
These look good and make sense, thanks for the help on this!

PR #483, opened 2022-04-06
Add delete confirmation to placements

edwardhorsford:
Adds a delete confirmation to placements, similar to #481

Screenshot 2022-04-06 at 13 10 54


edwardhorsford:
[Prototype: add delete confirmation to placements](https://trello.com/c/tiQ48wvZ/3969-prototype-add-delete-confirmation-to-placements)

ralph-hawkins (approved):
Given we're using the basically same page for all the delete confirmations (inc degrees), should that be a template?

PR #486, opened 2022-04-08
Misc app updates

edwardhorsford:
Key changes:
* Logs out the working record to the browser console
* Removes unused feature flags
* Swaps the bulk feature flags to use a radio.


app/views/_templates/_new-record.html
```diff
@@ -44,6 +44,8 @@
 
 {% block content %}
 
+  {{ record | log(record | getShortName + ":") }}
```
ralph-hawkins: - can we set an else if there's not short name please?
- i don't think you need a colon as it get rendered with a colon anyway, so you get two.

ralph-hawkins (approved):

app/views/_templates/_new-record.html
edwardhorsford: Good idea! fixed.

PR #487, opened 2022-04-08
Remove guard question from diversity section

edwardhorsford:
This removes the guard question from the diversity section. This mirrors work that happened on Apply last year.

In most cases, one or both of the diversity questions is answered - so having a guard question usually adds extra steps.

As a byproduct, the summary card will now always have rows for ethnicity and disabilities.


edwardhorsford:
[Prototype: remove guard question from diversity section](https://trello.com/c/ZpmWjAV6/3978-prototype-remove-guard-question-from-diversity-section)

ralph-hawkins (approved):

PR #493, opened 2022-04-14
Bulk update move data to routes

ralph-hawkins:


app/lib/utils.js
```diff
@@ -1720,8 +1722,9 @@ exports.filterByCannotBulkUpdate = (records) => {
 }
 
 // Trainees that can be bulk recommended
-exports.filterByCanBeRecommended = (records) => {
```
edwardhorsford, line 1723: I wonder if this filter could simply:
* Check it's `TRN received`
* Check it's complete

I think it could also be general named more general purpose?

app/routes/bulk-update-routes.js
```diff
@@ -39,9 +45,67 @@ module.exports = router => {
...
+    let templateErrors = [
+      'TRN not recognised',
+      'TRN missing',
+      'Trainee start date: ‘07/20/2023’ — enter a valid start date',
+      'Trainee start date: ‘20/07/2023’ — trainee start date must be in the past',
+      'URN not recognised',
+      'school is closed'
+    ]
+
+    let processedRows = uploadedTrainees.map((trainee, index) => {
```
edwardhorsford, line 68: Could you add a brief comment for this function? Something like `For each record, randomly pick whether it's ok, in error, or unchanged. If in error, pick a random error`

app/routes/bulk-update-routes.js
```diff
@@ -39,9 +45,67 @@ module.exports = router => {
...
+      }
+      return row
+    })
+
+    data.bulkUpload = {
+      processedRows
+    }
+
+    const checkIfAnyErrors  = element => element.uploadStatus == "error"
+    const checkIfAnyUpdated = element => element.uploadStatus == "updated"
```
edwardhorsford, line 99: I don't think these are named correctly. They don't check for 'any' errors - they check if a single thing has an error. If you want to check any, this function should include the `.some()`

edwardhorsford (changes requested):
Minor comments in code, which could be addressed later.

The csv you get when you first go down the recommend flow is broken - it's including stuff about errors and all squashed in to one cell.

This file seems oddly named?
Screenshot 2022-04-20 at 10 36 42

app/lib/utils.js
ralph-hawkins: True

edwardhorsford (approved):

PR #495, opened 2022-04-21
Set HEI as default, rename to fake HIE, increase numbers

edwardhorsford:
This:
* Sets the default org to an HEI for user research
* Renames our default org to a fictitious uni `Kings Oak University`
* Greatly increases the number of trainees at Kings Oak

As a compromise for performance, I've limited the number of historic years of trainees we'll generate for Kings Oak.


app/data/direct-add-draft-record.json
```diff
@@ -4,7 +4,7 @@
     "items": []
   },
   "reference": "KN5837",
-  "provider": "Coventry University",
+  "provider": "Kings Oak Unviersity",
```
ralph-hawkins: Does this need to match?

ralph-hawkins (approved):

app/data/direct-add-draft-record.json
edwardhorsford: Ah good catch. A typo in `university` hence my find and replace failing.

PR #502, opened 2022-04-25
Improved content to have clear steps

MaeveRoseveare:
We had feedback that this page could be clearer. It also did not align with the comms that are going out which sets out the steps in a clearer way.

I've tweaked the content to add similar steps e.g. Step 1: do this, Step 2: do this.

I think I've got the headings and their classes right but worth checking.

Old page:
Request an account old

New page:
Request an account new

### DIFF
--- a/app/views/request-an-account.html
+++ b/app/views/request-an-account.html
@@ -34,8 +34,8 @@
     <h2 class="govuk-heading-m">Step 3: Start using Register</h2>
     <p class="govuk-body">Once you’ve sent us the email in Step 2, we will set up your Register account.</p>
     <p class="govuk-body">We will then email you a link to start using Register.</p>
-    <h2 class="govuk-heading-m">Step 4 (optional): Access to multiple providers or lead school organisations</h2>
-    <p class="govuk-body">If you need access to multiple providers or lead school organisations, you do not need to create multiple user accounts.</p>
+    <h2 class="govuk-heading-m">Step 4 (optional): Access to multiple providers or lead schools</h2>
+    <p class="govuk-body">If you need access to multiple providers or lead schools, you do not need to create multiple user accounts.</p>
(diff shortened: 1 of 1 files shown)


app/views/request-an-account.html
```diff
@@ -13,22 +13,30 @@
...
       <li>the email address used for your DfE Sign-in account</li>
     </ul>
-    <p class="govuk-body">We will email you once we have set up your account.</p>
-
+    <p class="govuk-body">If there is more than one person at your organisation who would like to get set up, you can send one email with everyone’s details.</p>
+    <h2 class="govuk-heading-m">Step 3: Start using Register</h2>
+    <p class="govuk-body">Once you’ve sent us the email in Step 2, we will set up your Register account.</p>
+    <p class="govuk-body">We will then email you a link to start using Register.</p>
+    <h2 class="govuk-heading-m">Step 4 (optional): Access to multiple providers or lead school organisations</h2>
+    <p class="govuk-body">If you need access to multiple providers or lead school organisations, you do not need to create multiple user accounts.</p>
```
edwardhorsford: I think this should either be
`multiple providers or lead schools` or `multiple organisations`

app/views/request-an-account.html
```diff
@@ -13,22 +13,30 @@
...
       <li>that you would like to start using Register</li>
       <li>the email address used for your DfE Sign-in account</li>
     </ul>
-    <p class="govuk-body">We will email you once we have set up your account.</p>
-
+    <p class="govuk-body">If there is more than one person at your organisation who would like to get set up, you can send one email with everyone’s details.</p>
+    <h2 class="govuk-heading-m">Step 3: Start using Register</h2>
+    <p class="govuk-body">Once you’ve sent us the email in Step 2, we will set up your Register account.</p>
+    <p class="govuk-body">We will then email you a link to start using Register.</p>
+    <h2 class="govuk-heading-m">Step 4 (optional): Access to multiple providers or lead school organisations</h2>
```
edwardhorsford: I think this should either be
`multiple providers or lead schools` or `multiple organisations`

edwardhorsford (approved):

PR #504, opened 2022-04-26
makes smart quotes dumb

ralph-hawkins:
Smart quotes look aren't encoding correctly for Microsoft excel users, so switching to dumb quotes.


edwardhorsford (approved):
😢 

Can you add a description to the PR for why you're making this change?

PR #507, opened 2022-04-26
Added content about the new domain

MaeveRoseveare:
Unsure how to unlink the URL as this will be on the homepage.

### DIFF
--- a/app/data/service-updates.yaml
+++ b/app/data/service-updates.yaml
@@ -3,12 +3,12 @@
 items:
 
   - date: '2022-04-27'
-    title: Register’s website address has changed
+    title: Register’s website address (URL) has changed
     content: >-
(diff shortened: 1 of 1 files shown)


app/data/service-updates.yaml
```diff
@@ -2,6 +2,15 @@
 
 items:
 
+  - date: '2022-04-27'
+    title: Register’s website address has changed
+    content: >-
+      The website address for Register has now changed to: www.register-trainee-teachers.service.gov.uk
```
ralph-hawkins:
```suggestion
      The website address for Register has now changed to: <span class="app-nowrap">www&period;register-trainee-teachers.service.gov.uk</span>
```

Apparently you can use HTML reference character instead of a full stop. Also I think we should have a no wrap class.

ralph-hawkins (approved):

PR #508, opened 2022-04-27
Adds provider’s trainee ID to bulk recommend

ralph-hawkins:
We've seen in the first two rounds of research that users would find it difficult to identify trainees by their TRN alone.

To avoid needing to add names to the spreadsheets, we can give users the trainee ID. This probably isn't sufficient but it would be good to rule it out before reverting to needing to use the trainee names.

## Updated pre-populated file column descriptions

Screenshot 2022-04-27 at 12 28 15

## Updated template column descriptions

Screenshot 2022-04-27 at 12 32 20

## Update error page

- shows both IDs
- tweaks error message about missing ID
- adds error message about TRN and Trainee ID not matching

Screenshot 2022-04-27 at 12 25 24

## Updates pending changes

Shows IDs in their own column.

Screenshot 2022-04-27 at 12 40 59

### DIFF
--- a/app/assets/downloads/recommendOnly-template.csv
+++ b/app/assets/downloads/recommendOnly-template.csv
@@ -1,7 +1,8 @@
-TRN,Date standards met
+TRN,Trainee ID,Date standards met
 "Add the TRN for trainees you want to recommend.
+Can be blank if Trainee ID is provided","Your ID for the trainee.
 
-TRNs must be 7 digits.","Add the date when the trainee met QTS or EYTS standards.
+Can be blank if TRN is provided.","Add the date when the trainee met QTS or EYTS standards.
 Must be written DD/MM/YYYY. 
 For example, if the trainee met the teaching standard on 20 July 2022, write '20/07/2022'.
 The date must be in the past.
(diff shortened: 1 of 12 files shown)


ralph-hawkins:
Also adds trainee names to the csvs:

Screenshot 2022-04-29 at 13 50 38

ralph-hawkins:
Adds 'First names' and 'Last names' to the csvs

Screenshot 2022-04-29 at 14 01 05

app/assets/downloads/recommendOnly-template.csv
```diff
@@ -1,8 +1,8 @@
-TRN,Date standards met
-"Add the TRN for trainees you want to recommend.
-
-TRNs must be 7 digits.","Add the date when the trainee met QTS or EYTS standards.
+TRN,Trainee ID,Date standards met
+"Add the TRN for trainees you want to recommend.  Can be blank if Trainee ID is provided","Add the Trainee ID for trainees you want to recommend.
```
edwardhorsford: Should `Trainee ID` be capitalised like this?

ralph-hawkins:
Changed "Trainee ID" to "Provider ID" and made everywhere we show names "Last name, First name".

Error summary gets a lot busier, but I think its better to show all the relevant data.

Screenshot 2022-04-29 at 16 30 45

Check pending — have gone back to a single column for trainee name and both IDs

Screenshot 2022-04-29 at 16 13 06

edwardhorsford (approved):

app/assets/downloads/recommendOnly-template.csv
ralph-hawkins: no, was a hang over from using "Trainee ID", have made it consistent

PR #512, opened 2022-04-29
Improve display of missing data on records

edwardhorsford:
HESA records are sometimes missing data that we'd normally require. Since this can't be modified by users, we'll style this differently than our standard blue missing style.

New style:
Screenshot 2022-04-29 at 16 47 53

PR updates the seeds so that these items are commonly missing from HESA records:
* Nationality
* ITT End date

I also tweaked the placement summary card so that it will *always* show at least two rows. If you're missing both placements it only shows one call to action link though.

A record missing one placement:
Screenshot 2022-04-29 at 16 52 20

A record missing both placements:
Screenshot 2022-04-29 at 16 52 38

A HESA record missing one placement:
Screenshot 2022-04-29 at 16 54 13

A HESA record missing both placements:
Screenshot 2022-04-29 at 16 53 51

### DIFF
--- a/app/views/_includes/filter-panel/source.njk
+++ b/app/views/_includes/filter-panel/source.njk
@@ -22,11 +22,11 @@
       text: "Imported from DTTP",
       value: "DTTP",
       checked: checked(query.filterSource, "DTTP")
-    } if data.isAdmin and navActive !== "drafts",
+    } if navActive !== "drafts",
     {
       text: "Imported from HESA",
       value: "HESA",
       checked: checked(query.filterSource, "HESA")
-    } if data.isAdmin and navActive !== "drafts"
+    } if navActive !== "drafts"
   ]
 } | decorateAttributes(data, "data.filterSource"))}}
--- a/app/views/records.html
+++ b/app/views/records.html
@@ -63,7 +63,6 @@
       html: adminOnlyFilters
     }) }}
     {% else %}
-      {% include "_includes/filter-panel/source.njk" %}
 
   {% endif %}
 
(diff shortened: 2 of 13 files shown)


app/views/records.html
```diff
@@ -62,6 +62,8 @@
       classes: "app-status-box--filter-outdent",
       html: adminOnlyFilters
     }) }}
+    {% else %}
+      {% include "_includes/filter-panel/source.njk" %}
```
ralph-hawkins: I'm not sure this should be at the top filter for non-admin users?

app/views/_includes/filter-panel/source.njk
```diff
@@ -22,11 +22,11 @@
...
       value: "DTTP",
       checked: checked(query.filterSource, "DTTP")
-    } if data.isAdmin and navActive !== "drafts",
+    } if navActive !== "drafts",
     {
       text: "Imported from HESA",
       value: "HESA",
       checked: checked(query.filterSource, "HESA")
-    } if data.isAdmin and navActive !== "drafts"
+    } if navActive !== "drafts"
```
ralph-hawkins, line 30: Can we hide this from SCITTs?

ralph-hawkins (approved):

app/views/_includes/filter-panel/source.njk
edwardhorsford: If we only show if they have different sources, any reason they shouldn't get this filter?

(nb: a bit wary of too many differences between SCITTs and HEIs - I don't think we currently special case either of them?)

app/views/records.html
edwardhorsford: Yes probably. Will move lower.

app/views/_includes/filter-panel/source.njk
ralph-hawkins: We show the same filters to everyone in the prototype. So it feels weird to always show this to SCITTS, given they'll never have HESA records. 

We could make the filters only show for relevant sources, but we'd end up with the same thing.

edwardhorsford: Well we'd have apply vs manual for scitts.

PR #515, opened 2022-05-09
Removed domain notification and added HESA ones

MaeveRoseveare:
- removed domain notification as that is not live on production yet
- added in notification about HESA extension
- added notification about HESA import now working.

This will now match the order on production.

I do not know how to make both the HESA notifications persist like it is on production.

### DIFF
--- a/app/data/service-updates.yaml
+++ b/app/data/service-updates.yaml
@@ -8,7 +8,7 @@ items:
       The import from HESA to Register is working normally. This means if you’re a Higher Education Institute (HEI), you can use Register to see data that you have updated and uploaded to HESA. This will allow you to complete your April census update by 13 May.
 
 
-      If you identify any further issues with your data, contact [becomingateacher@digital.education.gov.uk](mailto:becomingateacher@digital.education.gov.uk?subject=HESA%20import%20issues%20in%20Register).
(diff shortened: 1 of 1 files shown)


app/data/service-updates.yaml
```diff
@@ -2,13 +2,22 @@
 
 items:
 
-  - date: '2022-04-27'
-    title: Register’s website address (URL) has changed
+  - date: '2022-05-09'
+    title: HESA records now visible in Register for April census update
```
edwardhorsford:
```suggestion
    title: HESA records now visible in Register for the April census update
```

app/data/service-updates.yaml
```diff
@@ -2,13 +2,22 @@
...
     content: >-
-      The website address (URL) for Register has now changed to: <span class="app-nowrap">www&period;register-trainee-teachers.service.gov.uk</span>
+      The import from HESA to Register is working normally. This means if you’re a Higher Education Institute (HEI), you can use Register to see data that you have updated and uploaded to HESA. This will allow you to complete your April census update by 13 May.
 
 
-      You will be automatically redirected to the new URL if you use the old one, so you will not see any disruption to your work or the service.
+      If you identify any further issues with your data, contact [becomingateacher@digital.education.gov.uk](mailto:becomingateacher@digital.education.gov.uk?subject=HESA%20import%20issues%20in%20Register).
```
edwardhorsford: Do we want to mention that placement data has been captured, but is not yet shown on records?

edwardhorsford (approved):

PR #516, opened 2022-05-11
Impersonate organisation feature

edwardhorsford:
This PR adds support for 'impersonating' a provider - for an admin to be able to see Register as a provider sees it. This is primarily to help with debugging user reports and checking things are rendering correctly. It's increasingly becoming needed because we've got lots of things that change depending on the provider type, the system state, etc. So what admins see is not necessarily what the user is seeing.

I've added it as a call to action on each record page:
Screenshot 2022-05-11 at 17 40 46

In addition to record pages, we should add this call to action to the admin section when viewing a single provider.

When viewing as a provider, you inherit the access of that provider. However any changes you make still attribute to yourself at DfE.

When impersonating a provider, we show this in the header:
Screenshot 2022-05-13 at 16 44 25

---

Other details:
* When viewing a trainee with a lead school, we let you view as that lead school too.
* I've refactored how admins are handled in the prototype so that they now work based on the signed in user's name being `System admin`
* Added `System admin` as a caption in a few places
* When signed in as an admin, you no longer see a random school's name in the top right.


edwardhorsford:
[Prototype: add impersonate provider feature](https://trello.com/c/ZvpdXcwB/4100-prototype-add-impersonate-provider-feature)

app/views/layout.html
```diff
@@ -166,16 +166,24 @@
...
+
+
 {% endblock %}
 
 {% block beforeContent %}
+
+{% if data.isImpersonating | falsify %}
+  {% include "_includes/impersonate-provider-cta.njk" %}
+{% endif %}
+
```
ralph-hawkins: This looks weird above the phase banner.

app/views/_includes/footer-settings.njk
```diff
@@ -2,7 +2,7 @@
   <div class="govuk-width-container">
     <ul class="govuk-footer__inline-list">
       <li class="govuk-footer__inline-list-item">
-        <a class="govuk-footer__link" href="/admin">
+        <a class="govuk-footer__link" href="/settings">
```
ralph-hawkins, line 5: can you update the link on /index.html as well please?

app/views/_components/admin-feature/template.njk
```diff
@@ -1,6 +1,8 @@
 <div class="app-status-box app-status-box--admin {{params.classes}}">
   <p class="govuk-body">
```
ralph-hawkins, line 2: Can you put the content in a 2/3rds div?

app/views/_components/admin-feature/template.njk
```diff
@@ -1,6 +1,8 @@
...
-    <strong class="govuk-tag govuk-tag--purple">Admin</strong>
+    <strong class="govuk-tag govuk-tag--purple">Admin</strong>{% if params.title %}
+      <span class="govuk-!-margin-left-2">{{params.title}}</span>
```
ralph-hawkins, line 4: Not a big issue, i think this would look cleaner as a paragraph.

ralph-hawkins:
Screenshot 2022-05-13 at 12 17 24

This takes up a lot of vertical space and the paragraph and buttons say the same thing. I wonder if we could loose the paragraph text? Maybe we could style the buttons as links so they're less prominent?

app/views/layout.html
edwardhorsford: Any suggestions? I've given it a bit of top margin now:
Screenshot 2022-05-13 at 13 14 48

edwardhorsford:
Revised double button scenario to combine the text:
Screenshot 2022-05-13 at 13 15 43

We might decide to only offer this from an admin page - in which case we wouldn't have a double provider scenario.

app/assets/sass/components/_admin-banner.scss
```diff
@@ -0,0 +1,30 @@
...
+  background: govuk-tint(govuk-colour("purple"), 90);
+}
+
+.app-admin-banner__container {
+  @include moj-width-container;
+
+  &:after {
+    content: '';
+    display: inline-block;
+    width: 100%;
```
ralph-hawkins: I don't think you need the `width: 100%`?

ralph-hawkins (approved):
💥

PR #517, opened 2022-05-12
Fix invalid answer styling on autocompletes

edwardhorsford:
Fixes broken styling on the autocompletes when there were invalid answers.

A few different issues:
* When an item is in error, an input gets a second id in `aria-describedby` - the autocomplete js code assumed there was only one
* The right colours weren't being set.
* The improved autocomplete component macro wasn't applying classes it was passed.


ralph-hawkins (approved):
super

PR #521, opened 2022-05-18
Incident on wrong user getting access to Register - update request-an-account.html

MaeveRoseveare:
1. Added in section about 'Who can get an account'

For the first bullet - would a school who does placements ever think that because they offer placements they are an ITT provider?

2. edited bullet to say 'training provider you work for' instead of 'you represent'

### DIFF
--- a/app/views/request-an-account.html
+++ b/app/views/request-an-account.html
@@ -14,11 +14,11 @@
     <h1 class="govuk-heading-l govuk-!-margin-bottom-9">{{pageHeading}}</h1>
 
     <h2 class="govuk-heading-m">Who can get a Register account</h2>
-    <p class="govuk-body">You can only get a Register account if you are:</p>
+    <p class="govuk-body">You can only get a Register account if you are one of the following:</p>
     <ul class="govuk-list govuk-list--bullet">
-  <li>an initial teacher training (ITT) accredited provider</li>
-  <li>a lead school involved in School Direct training routes</li>
-  </ul>
+      <li>an initial teacher training (ITT) accredited provider</li>
+      <li>a lead school involved in School Direct training routes</li>
+    </ul>
     <h2 class="govuk-heading-m">Steps to get an account</h2>
     <p class="govuk-body">Follow these steps to get an account to use Register.</p>
     <h2 class="govuk-heading-m">Step 1: Get a DfE Sign-in account</h2>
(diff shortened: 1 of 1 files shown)


app/views/request-an-account.html
```diff
@@ -5,15 +5,22 @@
...
   <div class="govuk-grid-column-two-thirds-from-desktop">
     <h1 class="govuk-heading-l govuk-!-margin-bottom-9">{{pageHeading}}</h1>
 
-    <p class="govuk-body">Follow these steps to get an account to use Register trainee teachers (Register).</p>
+    <h2 class="govuk-heading-m">Who can get a Register account</h2>
+    <p class="govuk-body">You can only get a Register account if you are:</p>
+    <ul class="govuk-list govuk-list--bullet">
+  <li>an initial teacher training (ITT) accredited provider</li>
+  <li>a lead school involved in School Direct training routes</li>
+  </ul>
```
edwardhorsford: Indentation needs fixing on these

edwardhorsford: Would it be worth linking to anything for each of these?

app/views/request-an-account.html
```diff
@@ -5,15 +5,22 @@
...
 
 {% block content %}
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds-from-desktop">
     <h1 class="govuk-heading-l govuk-!-margin-bottom-9">{{pageHeading}}</h1>
 
-    <p class="govuk-body">Follow these steps to get an account to use Register trainee teachers (Register).</p>
+    <h2 class="govuk-heading-m">Who can get a Register account</h2>
+    <p class="govuk-body">You can only get a Register account if you are:</p>
```
edwardhorsford: could this imply you need to be both?

app/views/request-an-account.html
```diff
@@ -26,7 +33,7 @@ <h2 class="govuk-heading-m">Step 2: Email us to request a Register account</h2>
     <p class="govuk-body">Once you have a DfE Sign-in account with a named email address, email <a href="mailto:becomingateacher@digital.education.gov.uk?subject=Request a Register account" class="govuk-link">becomingateacher@digital.education.gov.uk</a></p>
     <p class="govuk-body">In your email, tell us:</p>
     <ul class="govuk-list govuk-list--bullet">
-      <li>the name of the training provider or lead school you represent</li>
+      <li>the name of the training provider or lead school you work for</li>
```
edwardhorsford: Should this refer to accredited provider?

edwardhorsford (changes requested):

app/views/request-an-account.html
MaeveRoseveare: For accredited provider we could link to https://www.gov.uk/guidance/initial-teacher-training-itt-accreditation or https://www.gov.uk/government/collections/initial-teacher-training
I feel like the first link might be better as it explains there's a process to be an accredited provider, it's not just anyone

For lead school possibly: https://www.gov.uk/guidance/school-direct-guidance-for-lead-schools

edwardhorsford (approved):

PR #523, opened 2022-05-23
Revised homepage, cohorts and ways of filtering your trainees

edwardhorsford:
This PR takes a second look at the cohorts and the ways we can help our users see their current and past trainees.

The cohorts feature has been hard to build because the trainee data is more patchy than anticipated. In addition, the way it's presented on the homepage has caused some confusion in user research.

This PR replaced the concept of cohorts for a different one - that we stop identifying your 'current' trainees (which included everyone still training from all years and anyone who started or finished this year) and instead just have 'in training' which would be any trainee with a status of pending, trn received, or recommended. We can have a separate filter to let you get to your awarded trainees from this year - but we no longer make an attempt to slice by 'current' year.

Other changes:
* Introduces the ability to filter by end year
* Changes the homepage to remove confusing links
* Changes status cards on homepage to not be about statuses but instead 'useful' groups of trainees that we think providers will commonly want to filter on
* Lots of under the hood code changes to enable filtering by end year.

## End years

For most trainees we have an ITT end date. We can use this to derive the academic year in which you end. This isn't possible for some HESA records though. For these records we use the duration of the course to derive the end academic year.

Note where a trainee has deferred and reinstated we don't know for sure what year they fall in - in future work we should collect an end date when reinstating.

----

Homepage
Screenshot 2022-05-23 at 15 03 50

Records list
Screenshot 2022-05-23 at 15 04 05


app/lib/utils.js
```diff
@@ -1571,6 +1575,9 @@ exports.filterRecords = (records, data, filters = {}) => {
       if (filters.trainingStatus.includes(record.status)) {
         return true
       }
+      else if (filters.trainingStatus.includes("Course not yet started") && exports.ittInTheFuture(record)){
```
MaeveRoseveare:
```suggestion
      else if (filters.trainingStatus.includes("Course not started yet") && exports.ittInTheFuture(record)){
```

MaeveRoseveare (approved):
Just a small tweak to the not started yet tile. Otherwise, as discussed think this is good to go.

PR #529, opened 2022-06-06
Update English as a second language

edwardhorsford:
The specialism `English as a second language` should have the allocation subject of `English` rather than `Modern languages`


edwardhorsford:
[Prototype: update English as a second language](https://trello.com/c/snxcmHGS/4214-prototype-update-english-as-a-second-language)

PR #533, opened 2022-06-23
Add reports section

edwardhorsford:
Adds a new section to the service where users can access reports. We can use this if we need to provide any arbitrary data or pages on specific topics.

## Added to the header:
Screenshot 2022-06-29 at 10 19 04

## Reports index:
Screenshot 2022-06-29 at 10 19 22

## Census report:

We've gone through the census methodology to pick out the things we should be filtering. We've got some further analysis to do on what makes a trainee self-funded and whether we can identify those trainees.

Screenshot 2022-06-29 at 10 19 48

## HESA report:

Our data from hesa only tells us the last time a record was updated, not when it was created or last submitted. This limits the filters we can do.

To start, I've suggested all updates since the start of the academic year, and 'recent' updates - suggesting 2 months.

We've put in a change request with HESA to request created date and last submitted date. This could allow us to revisit these reports.

Screenshot 2022-06-29 at 10 20 21

### DIFF
--- a/app/views/reports/hesa.html
+++ b/app/views/reports/hesa.html
@@ -11,11 +11,19 @@
     <h1 class="govuk-heading-l">{{pageHeading}}</h1>
 
     <h2 class="govuk-heading-m">
-      About the export
+      About this export
     </h2>
 
     <p class="govuk-body">
-      Use this export if you’d like to compare data provided to HESA with what Register stores. <a href="#" class="govuk-link">How HESA data maps to a trainee record</a> details the fields we use and how we convert trainee data.
+      Use these exports to compare data provided to HESA and how that data is stored in Register.
+    </p>
+
+    <p class="govuk-body">
+      The export will filter your trainee records with the field ‘hesa_updated_at’, which shows you the last time any trainee record was updated in HESA. The export removes columns where HESA does not provide data, for example home address.
+    </p>
+
+    <p class="govuk-body">
+      <a href="#" class="govuk-link">Our guidance on how HESA data compares to Register data</a> will give you a breakdown of how we convert HESA data into Register.
     </p>
 
     <p class="govuk-body">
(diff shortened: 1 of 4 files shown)


app/views/reports/hesa.html
```diff
@@ -0,0 +1,84 @@
...
+    <p class="govuk-body">
+      The export will filter your trainee records with the field ‘hesa_updated_at’, which shows you the last time any trainee record was updated in HESA. The export removes columns where HESA does not provide data, for example home address.
+    </p>
+
+    <p class="govuk-body">
+      <a href="#" class="govuk-link">Our guidance on how HESA data compares to Register data</a> will give you a breakdown of how we import HESA data into Register.
+    </p>
+
+    <p class="govuk-body">
+      This exports your trainee records, filtering trainees using the fields ‘hesa_updated_at’{#  or ‘hesa_last_uploaded_at’ #}.
```
MaeveRoseveare: Think this paragraph is the old one so should be removed?

app/views/reports/hesa.html
```diff
@@ -0,0 +1,84 @@
...
+          classes: "govuk-fieldset__legend--m"
+        }
+      },
+      hint: {
+        text: ""
+      },
+      items: [
+        {
+          value: "2021 to 2022 cycle",
+          text: "Trainee records updated in HESA in the " + academicYear + " academic year (500 records)"
```
MaeveRoseveare, line 48: In the deployment the academic year is not reflecting it says "Trainee records updated in HESA in the undefined academic year"

MaeveRoseveare (commented):
Just spotted a few things that maybe are issues with the code?

app/views/reports/hesa.html
edwardhorsford: woops my bad

app/views/reports/hesa.html
edwardhorsford: Good catch - fixed now.

PR #541, opened 2022-07-20
Let admins change accrediting provider

edwardhorsford:
Adds the ability for admins to change the accrediting provider of a trainee.

When signed in as an admin, users get a 'change' link on the accrediting provider row, leading them to a new flow.

Trainee summary card:
Screenshot 2022-07-20 at 15 05 56

Edit page:
Screenshot 2022-07-20 at 17 40 26

Confirm page:
Screenshot 2022-07-20 at 17 40 34

Success banner:
Screenshot 2022-07-20 at 15 06 36

To test:

Go to prototype settings and turn on admin mode. From there, navigate to a trainee and change their provider.

### DIFF
--- a/app/views/_includes/forms/support/change-accrediting-provider.html
+++ b/app/views/_includes/forms/support/change-accrediting-provider.html
@@ -1,0 +1,69 @@
+{% set providers = data.signedInProviders %}
+
+{% set providers = data.providers.accreditingProviders.all | sort(attribute = "name") %}
+
+{% set providerItems = [] %}
+
+{% for provider in providers %}
+  {% set providerItems = providerItems | push({
+    text: provider.name + " (" + (provider.providerCode or "C23") + ")",
+    value: provider.name
+  }) %}
+{% endfor %}
+
+{# {% set providerItems = providerItems | push({divider: 'or'}) %}
+ #}
+
+<h1 class="govuk-heading-l">{{pageHeading}}</h1>
+
+{% set adminInsetContentHtml %}
+  <p class="govuk-body">
+    This feature is only available to admins.
+  </p>
+  <p class="govuk-body">
+    The old and new providers will both be sent a notification that the accrediting provider for this trainee has been updated.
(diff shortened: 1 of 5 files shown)


edwardhorsford:
[Prototype: trainees transferring from Org A to Org B](https://trello.com/c/7tQ8FQET/4413-prototype-trainees-transferring-from-org-a-to-org-b)

claire-hughez:
Screenshot 2022-07-20 at 17 14 20
So, should it go blank once you click on it as opposed to still having the provider name? 

It's useful to know what the old provider was, but when I click on it I can't see any others, till I backspace whats there. Just wondering.

I wonder if we need to flag any content to make it very obvious to Support that this could lead to a data breach too, if the provider is changed to an incorrect provider.

I could see them trying to change a lead school, and getting confused and changing the accredited provider instead for example.

edwardhorsford:
Made a few changes:

* Autocomplete does not pre-fill the existing provider as they're fiddly to change values on. 
* I've included the current provider's name in the intro content.
* Warning added on the confirmation page.

claire-hughez:
This looks great. Do we need to have a ticket to write the content for the notification?

Also, does Register have automated notifications yet (I think not but just checkin'), or is it something we should ask Support to send in these instances?


Screenshot 2022-07-21 at 10 43 14

Sorry if this is a dumb question, but is the text... spacing, meant to be tighter here? Is that a design pattern we have or something?

edwardhorsford:
> Also, does Register have automated notifications yet (I think not but just checkin'), or is it something we should ask Support to send in these instances?

We currently send one automated one - our welcome email.

We don't _have_ to send one, but to me it seems sensible to tell both providers (though who at the provider would we notify? team inbox time!). If we go with it (to discuss with devs?) yeah we'd want a content ticket.

claire-hughez:
I think MVP we would want to get just Support to notify both providers. They will know the user that requested the change and who they approved it with.

PR #555, opened 2022-08-19
Add section to compare HESA data with Register data

edwardhorsford:
Adds a new section with content from @MaeveRoseveare to compare HESA fields with their corresponding Register fields and column names.

Screenshot 2022-08-30 at 14 37 09


edwardhorsford:
[Update prototype with content for mapping between HESA and Register](https://trello.com/c/xLr8JcmV/4570-update-prototype-with-content-for-mapping-between-hesa-and-register)

PR #556, opened 2022-08-23
Refer to expected end dates

edwardhorsford:
Change this field to refer to `Expected end date` rather than `ITT end date`.

This is because we want users to update this if they know the trainee is likely to finish late, and to match the data we'll now be getting from HESA this academic year.


edwardhorsford:
[Refer to estimated itt end date](https://trello.com/c/XFel6QgB/4585-refer-to-estimated-itt-end-date)

MaeveRoseveare (approved):

PR #558, opened 2022-08-24
Ask for updated estimated end date when reinstating a trainee

edwardhorsford:
When a trainee reinstates, it's likely that their existing estimated end date is incorrect. They may have deferred for 3 years so their existing end date could have long since passed.

This adds a new page shown as part of the reinstate flow, asking the user to give us a new estimated end date.

I've done two variations of design for this, a simpler one, and a more complex one. For now, we're likely to go with the simpler one so that's what's default. You can edit the code to try the more complex one.

## Simpler design.

User is shown a date input with appropriate dynamic hint text:
Screenshot 2022-08-30 at 12 36 09

## More complex design

User is shown a radio question asking if the estimated end date has changed:

Screenshot 2022-08-30 at 12 37 40
Screenshot 2022-08-30 at 12 37 46

If the reinstate date is after the existing expected end date, or the expected end date is in the past, then we'd bypass the radio and show a date input:
Screenshot 2022-08-24 at 15 58 56

## Summary / check answers page

A new row is added for the expected end date

Screenshot 2022-08-24 at 16 11 06


edwardhorsford:
[When a trainee is reinstated from a deferral, we should ask their ITT estimated end date](https://trello.com/c/muwYl1Bb/4598-when-a-trainee-is-reinstated-from-a-deferral-we-should-ask-their-itt-estimated-end-date)

app/views/_includes/forms/reinstate/update-end-date.html
```diff
@@ -0,0 +1,88 @@
...
+{# Set to false to enable alternate flow where we ask a radio question about change #}
+{% set alwaysShowNewExpectedEndDate = true %}
+
+{% set startWithDateInput = reinstateIsAfterEndDate or expectedEndDateIsInThePast or alwaysShowNewExpectedEndDate %}
+
+{% include "_includes/trainee-name-caption.njk" %}
+
+{% set dateHint -%}
+  {% if alwaysShowNewExpectedEndDate or expectedEndDateIsInThePast -%}
+    They were expected to end on {{record.courseDetails.endDate | toDateArray | join('/') }} before they deferred. If this is still the same, you can enter that date again.
```
MaeveRoseveare:
```suggestion
    They were expected to end their course on {{record.courseDetails.endDate | toDateArray | join('/') }} before they deferred. If this is still the same, you can enter that date again.
```

MaeveRoseveare (approved):

PR #570, opened 2022-09-15
Add training years, filtering by training year and new export flow

edwardhorsford:
This PR adds support for the concept of training years - an attribute that tracks the academic years a trainee was in ITT.

We can use this to get all trainees in training this year, or that were in training last year.

Note: I had to tweak quite a bit of the seed record generation as it wasn't generating very realistic records crossing multiple years.

With this new data we can do some new things:

* New home page card for trainees from this year
* New filters to filter by training year
* New export flow for the reports section
* Link to export flow if a user does not have any filters applied

## Example for how training years are calculated

A trainee that started in 2019 to 2020 and is still in training in 2022 to 2023 will have training years of:
`['2019 to 2020','2020 to 2021', '2021 to 2022', '2022 to 2023']`

For finished trainees, it'll be all the years between start year and end year. For trainees still in training, it'll be the years from start year to the current year.

## New home page tile

Screenshot 2022-09-30 at 15 52 55

## New filters and linking to export flow when no filters applied

Screenshot 2022-09-30 at 15 53 35

### Reports index

Screenshot 2022-09-30 at 15 54 11

## Page 1 - asking which year
Screenshot 2022-09-30 at 15 54 33

## Page 2a - asking if they want all trainees or by start or end year
Screenshot 2022-09-30 at 15 55 02

## Page 2b - alternate if previous year
Screenshot 2022-09-30 at 15 55 27

## Page 3 - statuses and sources
Screenshot 2022-09-30 at 15 55 50

## Success
Screenshot 2022-09-30 at 15 56 10


edwardhorsford:
Rather than checkboxes for the filters, we also tried tabs:

Screenshot 2022-09-30 at 16 00 29

I think this could be good as something to explore in the future as a default view - most users will likely care about the current year.

The tabs have some downsides to work out though. Should search be scoped just to the tab? That might make it hard to find awarded trainees as users would likely search from the first tab. We could swap to the 'all years' tab when searching though.

You can swap to the tabs design using prototype settings.

PR #579, opened 2022-10-10
Add new degree grade options and grade autocomplete

edwardhorsford:
This adds two new default radio options for degree grades: distinction and merit. These are the most common other grades that we don't currently have options for.

Screenshot 2022-10-10 at 17 27 09

It also adds a suggestion autocomplete using the dfe reference data:
Screenshot 2022-10-10 at 17 29 46


edwardhorsford:
Also includes a bit of code cleanup and generating grades that are more realistic for higher qualifications.

PR #581, opened 2022-10-21
Add link to GIAS to help providers find schools

edwardhorsford:
We've had a couple cases of providers not able to locate schools - it would help if they searched first on GIAS before contacting us.

In one example, they were searching using a postcode that was different than the postcode in GIAS

In the other the school name was too common and they had the wrong URN.


MaeveRoseveare (approved):
Looks good, although the GIAS links don't work on both pages. I thought that might be intentional, but if not then now you know.

PR #582, opened 2022-10-24
Prevent admins exporting more than 100 trainees a go

edwardhorsford:
We're going to prevent admins exporting more than 100 trainees a go.

Screenshot 2022-10-24 at 12 07 21


MaeveRoseveare (approved):
Looks good!

PR #587, opened 2022-11-01
Clean up withdrawal data and move date to own page

edwardhorsford:
This moves the withdrawal date question to its own page. This matches how we collect dates in other flows and means the reason page will be able to grow if needed.

I also took the opportunity to refactor how the data was stored - now put in a `withdraw` object rather than directly on the record.

New withdrawal date page:
Screenshot 2022-11-02 at 15 50 23

Page if trainee is currently deferred:
Screenshot 2022-11-02 at 15 51 14


edwardhorsford:
[Prototype: clean up withdrawal flows](https://trello.com/c/Ut63JPDo/4958-prototype-clean-up-withdrawal-flows)

PR #592, opened 2022-11-15
Improve EBacc guidance

edwardhorsford:
We no longer need SCITT users to email us about EBacc trainees.

Collaboration with @johndoates


edwardhorsford:
[Change content on Register guidance pages around PE with Ebacc](https://trello.com/c/4zhInrUq/4989-change-content-on-register-guidance-pages-around-pe-with-ebacc)

claire-hughez:
Looks great

PR #595, opened 2022-11-16
Added placement guidance

johndoates:
We want providers to know that they will be asked about placements later in the academic year, so they should keep a record even though they can't add the data now.

CleanShot 2022-11-15 at 17 58 44


johndoates:
[Make 'Schools' wording on 'Check what data you need' guidance page for SCITTs clearer.](https://trello.com/c/F1C5wv0D/5000-make-schools-wording-on-check-what-data-you-need-guidance-page-for-scitts-clearer)

PR #611, opened 2022-12-15
Add extra tab for admins

edwardhorsford:
This adds an extra tab to trainee records that is only shown to admins.

The idea is that this tab can be used to contain extra metadata or options about a record that are only for admins.

To start, we can use it to access HESA data.

Screenshot 2022-12-15 at 17 19 13

### DIFF
--- a/app/views/_components/sub-navigation/template.njk
+++ b/app/views/_components/sub-navigation/template.njk
@@ -8,7 +8,7 @@
       {% if item %}
 
         <li class="app-sub-navigation__item">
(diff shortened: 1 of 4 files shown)


edwardhorsford:
[Design a way to solve the problem of us not being able to easily know the original XML submitted for a HESA trainee](https://trello.com/c/timFxh6A/5034-design-a-way-to-solve-the-problem-of-us-not-being-able-to-easily-know-the-original-xml-submitted-for-a-hesa-trainee)

PR #615, opened 2023-01-04
Update node, faker, weighted, uuid

edwardhorsford:
Several of these packages were several major versions behind.

Looks like faker changed their api for `date.between` so the dates all had to be provided in the reverse order.


edwardhorsford:
[Prototype: update dev dependencies](https://trello.com/c/t7v5Agc0/5102-prototype-update-dev-dependencies)

PR #617, opened 2023-01-13
Add end year and hesa id to trainee record, reorder rows

edwardhorsford:
Adds HESA ID and End academic year to the trainee record. Start year was already present. Also reorders the fields slightly.

Note the HESA ID row only shows if it's a HESA trainee.

This summary card is getting a bit long and busy - and we should probably think about splitting it in to two in the future.

Screenshot 2023-01-19 at 17 27 33

### DIFF
--- a/app/data/generators/hesa-data.js
+++ b/app/data/generators/hesa-data.js
@@ -17,9 +17,13 @@ module.exports = record => {
     record.courseDetails.endDate = weighted.select([null, endDate], percentageMissing)
   }
 
+  let hesaId = `2294839475${faker.datatype.number({'min': 1000000, 'max': 9999999})}, 1`
+
   delete record.contactDetails.address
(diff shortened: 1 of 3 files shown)


johndoates:
I suggest we leave Numhus out of the prototype for now, since we don't have it in Register yet.

Other than that, the change looks good. I agree that we should look at splitting it to make it easier to use.

edwardhorsford:
@johndoates updated without Numhus. Could you review again? thanks!

johndoates (approved):

edwardhorsford:
[Consider adding HESA_ID, record_source, Start_academic_year and End_academic_year to the Trainee record pages](https://trello.com/c/lrgNJjX6/5180-consider-adding-hesaid-recordsource-startacademicyear-and-endacademicyear-to-the-trainee-record-pages)

PR #618, opened 2023-01-31
Update performance profiles sign off pages to match production

johndoates:


edwardhorsford (changes requested):
I think this has accidentally included a bunch of changes that are coming as part of the bulk QTS work.

PR #620, opened 2023-02-06
Allow HESA records to be edited

edwardhorsford:
This adds a feature to let providers 'enable editing' of a HESA record.

We currently prevent editing of a HESA record because of the risk of data overwriting from HESA and otherwise getting out of sync. In most cases HEIs should be providing updates to us via HESA. However there are some cases where it it is either awkward or impossible to update via hesa:

Changes are needed and:
* a HESA update window has passed, but the trainee is ready to be awarded
* an academic year has just closed, and trainee has now withdrawn - the provider does not want to provide them in the next HESA collection purely to tell us they've left.
* the trainee was awarded or withdrawn more than a year ago so is not in the current collection.

We propose to allow providers to make changes to HESA records as long as they're not in a final state (withdrawn or awarded). We're proposing to make this an 'opt-in' feature - providers have to enable editing on the record. We've decided this because we'd like the defaults for most edits to still come via HESA - this should only be used in exception, and likely at the end of a trainee's period of training. It also gives us an opportunity to warn the provider that if they make changes and then provide the trainee to HESA again, that HESA data will overwrite it.

## HESA trainee record that's locked for editing (default):
Screenshot 2023-02-08 at 16 15 02

## Interstitial page warning of the consequences
Screenshot 2023-02-08 at 16 15 26

## HESA trainee record that's unlocked for editing:
Screenshot 2023-02-08 at 16 15 36

## Timeline message:
Screenshot 2023-02-08 at 16 15 46

## Interstitial if deferring a trainee:
Screenshot 2023-02-08 at 16 15 14

## Interstitial if reinstating a trainee:
Screenshot 2023-02-08 at 16 19 32


edwardhorsford:
[Consider how we could let users edit HESA records](https://trello.com/c/9EMyS0jG/5179-consider-how-we-could-let-users-edit-hesa-records)

claire-hughez:
Where does this come from? Do we need it for the enable editing designs?

'Last updated from HESA on 6 February 2023 at 4:23pm'

My concern is I'm not sure if we still have a reliable date/time stamp for this?

Otherwise, looks great

edwardhorsford:
It's intended to show the `hesa_updated_at` column, that I think is a timestamp of the last time this record was updated or created on the HESA side.

PR #621, opened 2023-02-07
Update question text to be clear it's about the last day of training

edwardhorsford:
When we ask for the withdrawal date, we want the last date the trainee was in training. Normally providers give this fine - but where the trainee was previously deferred, we would want the date they started deferral. We try to auto-populate this where we can, but we don't always have a deferral date. In this situation, providers are sometimes providing the date the trainee told them they were withdrawing, not when they were last in training - this could be months or years different.

This attempts to improve the situation by adjusting the legend and adding a heading. We've also adjusted the third radio option:

Screenshot 2023-02-07 at 16 37 39

Note we avoided mentioning when they left the course, because for HEIs / undergrads it may be that they're still on their course - it's just that they've left QTS or EYTS.

On production the page design is older, so here's how it would look there:

Screenshot 2023-02-07 at 16 41 00

### DIFF
--- a/app/views/_includes/forms/withdraw/date.html
+++ b/app/views/_includes/forms/withdraw/date.html
@@ -12,7 +12,7 @@
     </p>
   {% endset %}
 
-  <h1 class="govuk-heading-l">Withdrawal date</h1>
+  <h1 class="govuk-heading-l">{{pageHeading}}</h1>
 
   {{ govukInsetText({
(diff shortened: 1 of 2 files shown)


edwardhorsford:
[Update withdrawal date text to be clearer](https://trello.com/c/Bw5XQEGl/5239-update-withdrawal-date-text-to-be-clearer)

PR #624, opened 2023-02-13
Add transfer as a withdrawal option

edwardhorsford:
Adds the withdrawal reason of `transferred to another accredited provider`. If we later make a combined remove flow we may remove this again.


johndoates (approved):
Looks good.

PR #631, opened 2023-03-17
Allow SCITT users to select individual Apply applications to import

edwardhorsford:
Adds the ability for SCITT users to selectively import applications from Apply that are not yet in recruited status.

The intention is that this is only used in exception - where the trainee needs to be registered, but for one reason or another, their conditions are not all met yet.

New link on homepage
Screenshot 2023-03-24 at 13 18 24

Application selection page
Screenshot 2023-03-29 at 16 31 26

Confirm page
Screenshot 2023-03-29 at 16 32 05

Flash message
Screenshot 2023-03-29 at 16 34 20

### DIFF
--- a/app/assets/sass/overrides/_status-grid.scss
+++ b/app/assets/sass/overrides/_status-grid.scss
@@ -11,6 +11,10 @@
     grid-auto-rows: 1fr;
   }
 
+.app-home-statuses--1-up {
+  grid-template-columns: 1fr;
+}
+
 .app-home-statuses--2-up {
   grid-template-columns: 1fr;
   @include govuk-media-query($from: tablet) {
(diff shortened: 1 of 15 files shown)


edwardhorsford:
[Solve the problem of some trainees needing to be registered before they have met all of the conditions for their course on Apply for teacher training](https://trello.com/c/sFVrkbk9/5322-solve-the-problem-of-some-trainees-needing-to-be-registered-before-they-have-met-all-of-the-conditions-for-their-course-on-apply)

claire-hughez:
_Imported applications become draft trainee records in this service._
Given the last paragraph was on Apply, is it a bit confusing to use 'this service' here?

_are for a course with a start date between the beginning of last month and the end of next month_
This feels a little bit confusing to me. So if you import today, it could be from the start of February or the end of April? Is there a clearer way we could state this? Could we maybe also be a bit more forgiving here with the dates - a trainee could start in August, but they only really register when they're finalising census records in October. Particularly apprenticeships there's a bit of an a grey area.

'Conditions' are not available over the API at the moment, if we have to get rid of this, is there anything else we think we need here?

Otherwise, looks great, thank you

PR #649, opened 2023-05-22
Show award date as separate row

edwardhorsford:
Adds a second row to our award summary card to show the award date in addition to the standards met date.

Whilst trainee is recommended:
Screenshot 2023-05-23 at 15 03 25

Once they're awarded teaching status:
Screenshot 2023-05-22 at 15 43 14

Note the prototype doesn't currently distinguish between standards met date and award date - it just has as single date - therefore for this card I've faked it by adding 1 to the existing day.

PR #650, opened 2023-05-24
Show award date field when trainee recommended

johndoates:
We recently made a change to [show the award date as a separate row](https://github.com/DFE-Digital/register-trainee-teachers-prototype/pull/649/files).

We've now found out that the QTS and EYTS award date will always be the same as the date the standards were met. This means there's no point showing both dates once the trainee has been awarding teaching status.

When the trainee has been recommended for QTS or EYTS, we'll say that they're waiting for their award. We'll also give the date when they were recommended for teaching status.

CleanShot 2023-05-24 at 16 20 11

Once the trainee has teaching status, we'll just show the date they were awarded it.

CleanShot 2023-05-24 at 16 20 30

### DIFF
--- a/app/views/_includes/summary-cards/qualification-details.html
+++ b/app/views/_includes/summary-cards/qualification-details.html
@@ -1,33 +1,13 @@
 
 {% set outcomeDate = record.qualificationDetails.outcomeDate | toDateObject %}
 
-{# The prototype doesn’t store an award date, so fake it by adding 1 to the outcomeDate #}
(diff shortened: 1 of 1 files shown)


johndoates:
[Sort out what date show on a record when a trainee is recommended for QTS or EYTS](https://trello.com/c/5ljnfqAW/5478-sort-out-what-date-show-on-a-record-when-a-trainee-is-recommended-for-qts-or-eyts)

edwardhorsford (approved):

PR #656, opened 2023-06-05
Collect language specialisms with order

edwardhorsford:
The current language picker page lets you pick between 2 and 3 languages. When it was designed, the order of languages didn't matter - so it'll just get saved alphabetically.

For 2023 to 2024, there are Scholarships available only for certain languages. This means the order of languages is important. You'll only the the scholarship if the *first* language is eligible.

This updates the picker page to replace the checkboxes with two or three selects / autocompletes, explicitly collecting the languages with an order.

Screenshot 2023-06-08 at 15 29 24

The options in the select are all the modern languages, sorted alphabetically. As with the previous checkboxes, we've removed shortened the names of all of them where they're not plural.

Screenshot 2023-06-08 at 15 29 52

Note: as previously, the number of languages you can pick depends on how many subjects were picked in Publish. If the course is simply 'Modern languages' or has n languages, users can pick up to 3 total languages. If the course had two subjects like 'Biology with modern languages' then they can only pick 2 languages.

Screenshot 2023-06-08 at 15 33 39


edwardhorsford:
[Funding 23/24: Update course picker UI to account for language scholarhips](https://trello.com/c/orh6w66P/5539-funding-23-24-update-course-picker-ui-to-account-for-language-scholarhips)

PR #697, opened 2024-04-12
MVP for ITT route selection redesign

andrewscrivener:
- New placeholder route added. Name and data details tbc
- New layout for the select route view created for feedback


andrewscrivener:
Existing route selection view
Existing route selection view

Revised route selection view
Revised route selection view

PR #720, opened 2024-07-16
Replace node-sass with sass package

simonwhatley:


andrewscrivener (approved):

andrewscrivener:
To confirm we're moving the prototype off of node-sass as it doesn’t build on Apple Silicon and is deprecated.

